PR #20, opened 2026-04-16
fix: make MidenClient background sync loop resilient to transient errors

revitteth:
## Summary
- **Retry `ClientBuilder::build()`** with exponential backoff (1s→60s cap) so the proxy survives starting before the node is ready
- **Restart loop**: if the background thread crashes for any reason, it automatically restarts (5s cooldown) instead of leaving the proxy as a zombie where `/health` returns 200 but every write fails with "receiver is closed"
- **Real health check**: `/health` now returns 503 `{"status":"degraded","reason":"node connection lost"}` when the background thread is down, giving k8s readiness probes a real signal
- **Exponential backoff in `sync()`**: replaces fixed 5s retry with 1s→60s capped backoff to avoid hammering the node during sustained outages
- **Resilient `wait_for_transaction_commit`**: retries connection errors during tx polling (3 retries per attempt) instead of failing immediately
- **Prometheus counters**: `miden_client_build_errors_total`, `miden_client_restarts_total`, `miden_sync_errors_total` (with `kind` label)

## Test plan
- [x] `cargo build` — compiles cleanly
- [x] `cargo test` — all 65 unit tests pass
- [x] `cargo clippy` — no warnings
- [x] `make e2e` — full E2E suite passes (L1→L2, L2→L1, dynamic ERC-20, GER decomposition, security, concurrency)

### DIFF
--- a/src/metrics.rs
+++ b/src/metrics.rs
@@ -7,4 +7,13 @@ pub fn init_metrics() {
     describe_counter!("bridge_outs_total", "Total bridge-out operations");
     describe_counter!("store_errors_total", "Total store operation errors");
     describe_histogram!("rpc_request_duration_seconds", "JSON-RPC request duration");
+    describe_counter!(
+        "miden_client_build_errors_total",
+        "Failed attempts to build Miden client connection"
+    );
+    describe_counter!(
+        "miden_client_restarts_total",
+        "Background thread restarts after crash"
+    );
+    describe_counter!("miden_sync_errors_total", "Sync errors by kind");
 }
--- a/src/miden_client.rs
+++ b/src/miden_client.rs
@@ -10,12 +10,22 @@ use std::error::Error;
 use std::path::PathBuf;
 use std::pin::Pin;
 use std::sync::Arc;
+use std::sync::atomic::{AtomicBool, Ordering};
 #[cfg(test)]
-use std::sync::atomic::{AtomicUsize, Ordering};
+use std::sync::atomic::AtomicUsize;
 use std::thread;
 use std::time::Duration;
 use tokio::sync::mpsc;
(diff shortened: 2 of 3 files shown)


revitteth:
## Test report

### Standard suite — all green on PR branch

| Suite | Result |
| --- | --- |
| `cargo check/test/clippy/fmt` | clean, 75/75 |
| `e2e-test.sh` | 74 PASS / 0 FAIL |

### Stress test — 60 rapid sequential deposits

Sent 60 `bridgeAsset` calls to L1 at ~1s intervals to stress the GER injection path:

- **60/60 deposits sent** successfully
- **6 GER commits** (rest were deduped — same global exit root)
- **0 errors, 0 crashes, 0 "receiver is closed"** — proxy healthy throughout
- **Block progress: 204** — the original crash was at block 1452

### Reproduction status: not triggered

The `BlockHeaderNotFound(1452)` crash could not be reproduced in our local docker-compose stack. The original environment likely ran for much longer or the node's block-header storage/pruning behaved differently. Our local miden-node reached block 204 after the full test — well short of the ~1452 blocks needed.

**Why the fix is still correct regardless:** the crash path is a `?` operator on line 308 that propagates any `on_sync` error out of `run()`, killing the background thread. The fix replaces it with `if let Err(err) = ... { tracing::error!(...) }`. There is no code path where a sync/listener error can kill the thread now — the structural change is verifiable from the diff alone. The stress test confirms the proxy stays healthy under load.

### To fully reproduce

Would need either:
- A longer-running stack (10+ minutes of continuous deposits) where the node reaches block ~1400+
- Or artificially triggering `BlockHeaderNotFound` by pruning the miden-node's block storage mid-run

elliothllm (approved):

PR #32, opened 2026-04-23
fix(claim): scale claim amount via EthAmount, drop u32 cap

partylikeits1983:
## Summary

The claim service was rejecting valid L1 deposits before they could ever be claimed. For an 18 → 8 decimal token, any deposit above ~42.95 tokens failed outright, and any deposit with sub-unit wei also failed. Retries kept hitting the same check, so funds stayed stuck.

The cause was a local `validate_amount` helper that required the scaled amount to divide evenly and capped the result at `u32::MAX`. Both rules were service-local, not protocol rules. The library already ships `EthAmount::scale_to_token_amount`, which just floor-divides and enforces the real protocol ceiling (`FungibleAsset::MAX_AMOUNT`). This PR switches `create_claim` to use it and deletes the now-unused helper.

L1 state is untouched: `leaf_data.amount` still carries the full `EthAmount`, so leaf hashes and bridge proofs match exactly as before. Only the Miden-side payout amount changes, from "reject" to `floor(amount / 10^scale)`.

Fixes 0xMiden/protocol#2804.


revitteth (commented):
Reviewed end-to-end. Fix is sound — the old `validate_amount` imposed two service-local rules (exact-divisibility and `u32::MAX` ceiling) that were never protocol rules, and this replaces them with the actual protocol cap via `EthAmount::scale_to_token_amount`. Tests cover all four interesting cases. L1 `leaf_data.amount` still carries the full `EthAmount` so proofs and leaf hashes are unchanged — confirmed by reading the diff in `create_claim`.

## One substantive concern — please switch the `.expect()` to an `Err`

In `scale_claim_amount`:

```rust
let scale = u32::from(
    faucet
        .origin_token_decimals
        .checked_sub(faucet.decimals)
        .expect("find_or_create_faucet enforces origin_token_decimals >= decimals"),
);
```

`find_or_create_faucet` does enforce the invariant for auto-created faucets, but `Faucet` also gets built from existing `FaucetEntry` rows via `store.get_faucet_by_origin(...)`. If an operator ever seeds a row manually (or a migration goes wrong, or a future code path inserts without going through `find_or_create_faucet`), the invariant is only "enforced by convention". A panic in this code path is particularly bad: `publish_claim` runs inside `tokio::task::spawn_blocking` with a fresh `Runtime::new()`, so the panic aborts the worker and aggkit sees a malformed response rather than a clean error.

Suggestion — mirror the shape used in `find_or_create_faucet` itself:

```rust
let scale_byte = faucet
    .origin_token_decimals
    .checked_sub(faucet.decimals)
    .ok_or_else(|| {
        anyhow::anyhow!(
            "faucet {} has miden_decimals ({}) > origin_token_decimals ({}); \
             invariant violated, refusing to compute scale",
            faucet.id,
            faucet.decimals,
            faucet.origin_token_decimals,
        )
    })?;
let scale = u32::from(scale_byte);
```

Cheap belt-and-braces. One more regression test worth adding alongside:

```rust
#[test]
fn rejects_faucet_with_inverted_decimals() {
    // If a stored FaucetEntry somehow has miden_decimals > origin_token_decimals,
    // scale_claim_amount must error cleanly, not panic.
    let err = scale_claim_amount(&eth_amount(U256::from(1u64)), faucet(6, 8)).unwrap_err();
    assert!(format!("{err}").contains("invariant violated"));
}
```

## Non-blocking nits

1. **`Fixes 0xMiden/protocol#2804`** — worth double-checking the link in the PR body. The issue in the `0xMiden/protocol` repo may actually be ours (this repo) or in `miden-base-agglayer`. If 2804 is the right tracker, great; if not, quick swap.

2. **Silent truncation of sub-unit wei** — the behaviour change is correct (full value still carried in `leaf_data.amount`, only the Miden-side payout is floored), and the `truncates_sub_unit_wei_remainder` test documents it. Might be worth a one-line comment in `scale_claim_amount` pointing out that the full value is preserved on L1; a future reader asking "wait, we silently drop dust?" would appreciate it.

## Interaction with the in-flight PRs

- **PR #33** (publish_claim URL fix): touches `claim.rs` but in the fresh-client builder block. No real conflict with this PR; small rebase if #33 merges first.
- **PR #34** (swallow unresolvable-destination claims, RD-860): short-circuits **before** `create_claim` runs, so your scaling change never executes for swallowed claims. No conflict.

Happy to approve once the `.expect()` becomes an `Err` (or you push back with a good reason to keep it panicking — I can see the argument that a violated invariant means the DB is corrupt and we want a loud failure, just want it called out explicitly).

partylikeits1983:
Thanks for the thorough review. Addressed in 35f7ff2:

- `scale_claim_amount` now returns an `anyhow::Error` naming the faucet and both decimal values instead of `.expect(...)`ing. The message ends with `invariant violated, refusing to compute scale` so aggkit gets a clean response path instead of a `spawn_blocking` worker abort.
- Added the `rejects_faucet_with_inverted_decimals` regression test in the `mod scale_claim_amount` submodule.

On the L1-preservation nit: the existing doc comment on `scale_claim_amount` already calls this out (`the full value is still preserved in leaf_data.amount`), so leaving it as is unless you want it louder.

On the `Fixes 0xMiden/protocol#2804` link: the issue lives in the protocol repo, so GitHub won't auto-close it on merge. I'll close it manually with a pointer back here once this lands. Happy to open a mirror on this repo if you'd rather have the tracker local, just let me know.

revitteth (approved):
Approved: division logic verified against `EthAmount::scale_to_token_amount` (floor-divide + protocol `MAX_AMOUNT` ceiling). `leaf_data.amount` unchanged at claim.rs:180, so L1 leaf hashes / bridge proofs are unaffected. 5 new tests + full lib suite (74) passing locally on the PR HEAD.

PR #34, opened 2026-04-24
feat(claim): swallow unresolvable-destination claims (RD-860)

revitteth:
Fixes [RD-860](https://linear.app/gateway-fm/issue/RD-860/swallow-unresolvable-destination-claims-so-one-bad-claim-cant-wedge).

> ⚠️ **Stacked on #33** (`fix/dns-retry-claim-ger-build`). GitHub shows only the delta on top of #33. Merge #33 first, then this.

## Problem

If aggkit delivers a `claimAsset` whose `destinationAddress` can't be resolved to a Miden `AccountId` (not hardhat, not a store-registered mapping, not a zero-padded `MidenAccountId`), `publish_claim` bails with `no known Miden AccountId for Ethereum address …`. aggkit treats the `eth_sendRawTransaction` error as retryable and hammers the same failure forever. Impact:

1. Wedges aggkit's `claimtxmanager` in an infinite retry loop.
2. Potentially blocks later claims behind it (depends on aggkit's queue semantics).
3. Burns miden-agglayer CPU — every retry rebuilds a fresh `MidenClient` (per the `9ebcabb` state-drift workaround).

On a public bridge, this is a cheap griefing/DoS vector: attacker pays L1 gas + a dust deposit with a bogus destination and stalls the claim queue.

## Fix (tier 1)

In `service_send_raw_txn::handle_claim_asset`, pre-check `resolve_address` **before** `try_claim`:

- On success → existing flow, unchanged.
- On failure →
  - Record in new `unclaimable_claims` store table (for operator visibility + future rescue endpoint).
  - Emit the synthetic `ClaimEvent` so aggkit marks the globalIndex complete and stops retrying.
  - Return success to the caller.
  - Increment `claim_unclaimable_total{reason=unresolvable_destination}` metric.
  - `WARN` log with `global_index`, `destination`, `eth_tx`, `origin_*`, `amount`.

No L2 funds move; the original L1 deposit remains locked in the bridge contract. Recovery is operator-driven — see **tier 2** in the RD-860 ticket for a planned rescue endpoint.

## Security posture

| Attack goal | Before | After |
|---|---|---|
| Wedge aggkit's claim queue with one bad destination | Yes, trivially | No — single short-circuit returns success immediately |
| Steal L2 funds via unresolvable destination | No (no funds released) | No (no funds released) |
| Burn miden-agglayer CPU per retry | Yes, indefinitely | One-shot record + log |
| Hide attack from operators | No (errors logged, but buried in retry noise) | No — every swallow writes a durable row to `unclaimable_claims` with full context |

## Changes

- `migrations/003_unclaimable_claims.sql` — new table, column format mirrors the existing `claimed_indices` table so `unclaimable_claims ⋈ claimed_indices` joins cleanly on `global_index`.
- `src/store/mod.rs` — `UnclaimableClaim` struct, `UnclaimableReason` enum (currently only `UnresolvableDestination`, designed to extend), two new trait methods: `record_unclaimable_claim` (idempotent, returns `true` for newly inserted) and `get_unclaimable_claim`.
- `src/store/memory.rs` — `HashMap`-backed impls. Unit test: first-write-wins behaviour on duplicate `global_index`.
- `src/store/postgres.rs` — `INSERT … ON CONFLICT DO NOTHING RETURNING` for idempotent inserts; `SELECT` for reads.
- `src/service_send_raw_txn.rs` — short-circuit path ahead of `try_claim`. New test `test_claim_asset_unresolvable_destination_swallowed` covers: (1) no globalIndex lock, (2) `MidenClient` never invoked, (3) record lands in store with correct fields, (4) exactly one `ClaimEvent` emitted, (5) nonce incremented and receipt recorded.

**Note on `test_claim_asset_no_event_on_failure`:** it used `destinationAddress = [0x42; 20]`, which now hits the short-circuit. Updated to use a zero-padded resolvable destination (`0x000000003d7c9747558851900f8206226dfbea00`) so the test still exercises the `publish_claim`-error path it was designed for.

## Deliberately NOT in this PR

- **Tier 2 rescue endpoint.** Operator-facing `/admin/rescue-claim` that re-processes a stuck claim after registering a new destination mapping. Separate ticket.
- **Tier 3 auto-bridge-back-out.** Needs L1 bridge contract support for claim-failure refunds, which the Polygon agglayer bridge doesn't have as a first-class feature.

## Test plan

- [x] `cargo build` (default features)
- [x] `cargo build --features postgres`
- [x] `cargo test --lib` — 72 passed (was 70; +2 for swallow + first-write-wins)
- [x] `cargo clippy --features postgres --all-targets -- -D warnings`
- [x] `make lint` — rustfmt + taplo + typos + clippy all green
- [ ] Post-merge: a claim to any non-zero-padded address on testnet should now return an `eth_sendRawTransaction` OK + a row in `unclaimable_claims`, instead of the aggkit retry loop we saw during the RD-856 investigation.

## SQL for operators

```sql
-- current backlog
SELECT global_index, destination_address, amount, reason, created_at
FROM unclaimable_claims
ORDER BY created_at DESC LIMIT 20;

-- size of the backlog
SELECT reason, COUNT(*) FROM unclaimable_claims GROUP BY reason;
```

### DIFF
--- a/migrations/003_unclaimable_claims.sql
+++ b/migrations/003_unclaimable_claims.sql
@@ -1,0 +1,34 @@
+-- RD-860: log of L1→L2 claims we refused to process because the destination could
+-- not be resolved to a Miden AccountId.
+--
+-- Claims land here when `address_mapper::resolve_address` fails (not hardhat, no
+-- explicit store mapping, not a zero-padded MidenAccountId). We short-circuit them
+-- in `service_send_raw_txn::handle_claim_asset` by emitting a synthetic `ClaimEvent`
+-- so aggkit marks the globalIndex complete and stops retrying — WITHOUT actually
+-- minting any L2 funds. This row is the only record that the claim was effectively
+-- dropped; a future operator rescue endpoint (tier 2) can query this table.
+--
+-- `global_index` is the authoritative key from the L1 bridge. eth_tx_hash is the
+-- aggkit-submitted eth tx that carried the claim. Both are indexed for fast lookup
+-- when a user shows up asking "where did my deposit go?".
+
+-- Column conventions here mirror the existing `claimed_indices` table
+-- (migrations/001_initial.sql): U256 values stored as lowercase `0x`-prefixed hex
+-- TEXT, addresses / tx hashes as `0x`-prefixed TEXT. That lets us join and cross-
+-- reference the two tables without conversion.
+CREATE TABLE IF NOT EXISTS unclaimable_claims (
+    global_index        TEXT PRIMARY KEY,
+    destination_address TEXT NOT NULL,
+    origin_network      INT NOT NULL,
+    origin_address      TEXT NOT NULL,
+    amount              TEXT NOT NULL,
+    reason              TEXT NOT NULL,
+    eth_tx_hash         TEXT NOT NULL,
+    created_at          TIMESTAMPTZ NOT NULL DEFAULT now()
+);
+
+CREATE INDEX IF NOT EXISTS idx_unclaimable_claims_eth_tx_hash
+    ON unclaimable_claims (eth_tx_hash);
+
+CREATE INDEX IF NOT EXISTS idx_unclaimable_claims_destination
+    ON unclaimable_claims (destination_address);
--- a/src/service_send_raw_txn.rs
+++ b/src/service_send_raw_txn.rs
@@ -155,6 +155,72 @@ async fn handle_claim_asset(
         return Ok(txn_hash);
     }
 
+    // RD-860 — swallow unresolvable-destination claims permanently. If the
+    // destination address can't be resolved to a Miden AccountId, record the
+    // unclaimable entry, emit the synthetic ClaimEvent so aggkit marks the
+    // globalIndex complete and stops retrying, and return success to the
+    // caller. Funds remain locked on L1; an operator rescue endpoint (tier 2,
+    // future work) would let ops re-process by registering a destination
+    // mapping and replaying.
+    //
+    // Ordering: this check runs BEFORE C6's GER pre-check because the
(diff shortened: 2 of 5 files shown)


elliothllm (dismissed):

revitteth:
Refreshed by merging `main` (commit 275ce7e) — clears the stale 'Stacked on #33' state since #33 has long since merged. CI re-running.

Local validation against the merged tree: `cargo test --lib` 72 passing (incl. `test_claim_asset_unresolvable_destination_swallowed` and `test_claim_asset_no_event_on_failure`), `cargo clippy --features postgres --all-targets -- -D warnings` clean.

Aiming to include in v0.1.3 today alongside #32.

revitteth:
Rebased onto main (post-#38 merge) and resolved the conflict in `src/service_send_raw_txn.rs`. The two pre-`try_claim` checks coexist — RD-860 (unresolvable destination, permanent) runs first, then C6 (GER not seen, transient). RD-860 first because the destination state is permanent, so swallowing immediately avoids any GER-not-seen retry churn from aggkit. C6 test updated to use a resolvable zero-padded destination so the C6 path stays exercised. 141/141 lib tests green, clippy + fmt clean.

revitteth:
Rebased onto main (post-#38 merge) and resolved the conflict in `src/service_send_raw_txn.rs`. The two pre-`try_claim` checks coexist — RD-860 (unresolvable destination, permanent) runs first, then C6 (GER not seen, transient). RD-860 first because the destination state is permanent, so swallowing immediately avoids any GER-not-seen retry churn from aggkit. The C6 test was updated to use a resolvable zero-padded destination so the C6 path stays exercised. 141/141 lib tests green, clippy + fmt clean.

elliothllm (approved):

PR #44, opened 2026-05-18
v0.3.0: runtime self-heal + Public storage_mode + unified claim client

revitteth:
## ⚠️ Decisions pending user sign-off

Two items need your call before this fully closes:

### 1) `v0.3.0` tag location

A `v0.3.0` annotated tag was pushed to origin at commit `104beca` **before** the BLOCKING fix `4a5b3a2` was added. The tag is currently **missing the asymmetric-self-heal-coverage fix** that the pre-merge code review flagged as merge-blocking.

Three options:

| Option | Action | Pros | Cons |
|---|---|---|---|
| **A** (recommended) | Force-push `v0.3.0` tag to `4a5b3a2` | Single coherent release tag; no consumer confusion. Bounded blast radius — no docker image is published yet, no deployment has consumed the tag. | Force-push of an annotated tag is destructive per org policy; needs your explicit OK. |
| **B** | Leave `v0.3.0` at `104beca`; tag `v0.3.1` at `4a5b3a2` | No destructive ops. | Two release tags 5 minutes apart looks bad; bali redeploy must deploy `v0.3.1`, not `v0.3.0`, which contradicts every doc that references `v0.3.0`. |
| **C** | Delete `v0.3.0` from origin; re-tag at `4a5b3a2` | Same end-state as A. | Same destructive-op concern as A; intermediate state where no `v0.3.0` exists is bad if anyone has already pulled the old one. |

**My recommendation: A.** No consumers exist yet; the cost of force-push is zero in practice; the gain is one clean release tag. Please confirm before I execute.

### 2) Docker image build for `gatewayfm/miden-agglayer:0.3.0`

Held pending decision 1. Once the tag is at the right commit, the image build follows.

## What

v0.3.0 of `miden-agglayer-service`. Closes the chain of regressions that locked bali for ~20 days. Two structural fixes, three operator-recovery hardening fixes, plus indexer + startup + logging cleanups; before/after evidence captured against a fresh local stack; pre-merge security + functional review with one BLOCKING finding addressed.

**16 commits, +2,737 / -268 LoC across 24 files. Ship-ready after the BLOCKING fix at `4a5b3a2`.**

---

## Timeline — what happened when, on bali, in production

Reconstructed end-to-end from Loki (`outpost-testnet-miden-testnet/miden-agglayer-0`):

```
2026-04-13 14:41:57+0100   dbe5c2d  commit on `main`: 0.14.x miden-client migration
                                    silently drops `.storage_mode(AccountStorageMode::Network)`
[...]

PR #45, opened 2026-05-19
v0.4.0: runtime self-heal + Public storage_mode + unified claim client

revitteth:
## Summary

Supersedes #44 with a clean, hand-picked commit history.

Closes the bali production incident (Marti's L1→L2 deposit stuck ~20 days as of 2026-05-18) and structurally prevents the bug class.

## What's broken on bali (verified via Loki + code archaeology + local repro)

| When | Event |
|------|-------|
| 2026-04-13 | Commit `dbe5c2d` accidentally dropped `.storage_mode(AccountStorageMode::Network/Public)` from `add_wallet` during the 0.14.x migration. All accounts deployed from this point default to `Private` (no on-chain copy). |
| 2026-04-23 | Bali deployed via the buggy code path. `ger_manager`, `bridge`, `service` accounts all Private. |
| 2026-05-11 16:14:51Z | First `incorrect account initial commitment` (IAIC) log line on bali. Recurred 189 times through 2026-05-14 from concurrent claim + GER submissions racing the `bridge` account's mempool slot. |
| 2026-05-14 18:45:18Z | Operator ran `--reset-miden-store --restore` to recover. `--reset` wiped `store.sqlite3`; `--restore` Phase 1 only called `sync_state()` — does **not** re-import bridge accounts. `latest_account_headers` came back empty. |
| 2026-05-14 onwards | Every aggoracle `insertGlobalExitRoot` push fails with `AccountDataNotFound`. Backlog grows. |

### Cure for the active failure mode

- **`fix(init): use AccountStorageMode::Public for add_wallet`** — restores the dropped storage_mode call. Future deployments are recoverable from a local sqlite loss.
- **`feat(self-heal): runtime retry on AccountDataNotFound + IAIC`** — typed-downcast catch on `ClientError`/`AddTransactionError`, `import_account_by_id` from node, retry once. Covers both error types via the same code path. 5 unit tests pin the typed match against canonical upstream variants.
- **`feat(restore): import bridge accounts in Phase 0 of --restore`** — `--reset-miden-store --restore` now re-imports bridge accounts before `sync_state()`. End-to-end repro: `scripts/e2e-reset-restore-recovery.sh`.

### Structural prevention of the root-cause race

- **`feat(claim): route publish_claim through long-lived MidenClient`** — eliminates the dual-sqlite-handle race. Every Miden submission funnels through the long-lived client's `mpsc::channel(1)`. At most one in-flight submission ⇒ structurally impossible to produce two mempool-conflicting txs from the proxy.
- **`fix(claim): wrap publish_claim with self-heal retry`** — closes the asymmetry where only the GER path had the safety net.
- **`feat(indexer): persistent cursor + resume on restart`** — closes the gap that produced bali's 27 historic STATE-C orphans.

### Defensive hardening + operational fixes

- **`fix(logging): clamp alloy http transport to info`** — stops `RUST_LOG=debug` from logging the Sepolia archival `?apiKey=<SECRET>` query string in every L1 RPC line. Active SIEM credential leak.
- **`fix(accounts-config): atomic write for bridge_accounts.toml`** — fsync + rename. Prevents OOM-mid-write corruption.
- **`feat(startup): in-process DB migrator`** — embeds migrations via `include_str!` + checksum-verified `schema_migrations` + advisory lock. Replaces the `agglayer-migrate` compose service that was a deploy-time footgun (new migrations had to be wired into compose by hand).
- **`feat(indexer): --l1-indexer-from-block override`** — operator backfill flag for bali's 27 historic STATE-C orphans (cosmetic; non-blocking).
- **`fix(dockerfile): COPY migrations dir into builder context`** — without it the build fails at `include_str!`.

## Test surface

```
cargo test --lib --features postgres   →   179 passed; 1 ignored (live-PG)
cargo clippy --all-targets             →   0 warnings
cargo build --release --features postgres → 0 errors
```

E2E repro scripts (in `scripts/`):
- `e2e-account-self-heal.sh` (MODE=`expect_failure` / `expect_self_heal` — proves bug on main AND cure on this branch, same script)
- `e2e-reset-restore-recovery.sh` (operator-faithful reset+restore cure)
- `e2e-iaic-mempool-conflict.sh` (concurrent-load IAIC induction; requires fixture, see script header)

Multi-pass evidence from earlier runs and the regression sweep are not committed to the repo (kept local in `../repro/v0.3.0-evidence/`); happy to share out of band if you want to see them.

## Bali deployment caveat

v0.4.0's runtime self-heal **cannot** recover bali's current `ger_manager` because that account is Private (created pre-`dbe5c2d`-fix). `import_account_by_id` returns `AccountIsPrivate` on Private accounts. Bali specifically needs a **full redeploy** with fresh Public accounts + aggkit reconfig. Procedure documented in `docs/REDEPLOY_RUNBOOK_BALI.md`.

All FUTURE deployments (and any other proxy redeployed off v0.4.0) self-heal on this bug class without operator intervention.

## Why v0.4.0 and not v0.3.1?

The `v0.3.0` tag was pushed earlier today against a slightly different commit (`bb529f3`). Per org policy, force-updating a published tag is off the table. v0.4.0 sidesteps the ambiguity cleanly.

## Test plan
- [ ] Review the 13 substantive commits + 1 fmt commit
- [ ] Confirm bali redeploy procedure with SRE
- [ ] Merge → tag v0.4.0 → docker image push
- [ ] Bali redeploy (per `docs/REDEPLOY_RUNBOOK_BALI.md`)
- [ ] Aggkit reconfig pointing at new bridge account IDs
- [ ] Verify Marti's deposit `cnt=1130654` flips `ready_for_claim=true`

### DIFF
--- a/src/ger.rs
+++ b/src/ger.rs
@@ -94,6 +94,73 @@ pub struct GerInsertResult {
     pub is_new: bool,
 }
 
+/// Submit the actual UpdateGerNote Miden transaction. Factored out of
+/// `insert_ger` so the caller can run it twice — once eagerly, then again
+/// after `reimport_account` if the first attempt failed with a recoverable
+/// account-state error.
+///
+/// Use the long-lived MidenClient. The dedicated ger_manager account
+/// (separate from the service account that the NTX builder constantly
+/// mutates via claim processing) keeps the account state stable across
+/// GER submissions, so we don't need a fresh client per call.
+///
+/// Fresh-client-per-GER was removed because it shared the main sqlite
+/// and advanced the sync cursor past blocks where bridge NTX consumes
+/// the UpdateGerNote. The main client's subsequent sync_nullifiers only
+/// queries [current_cursor, tip], so those consumption events were never
+/// discovered and `NoteFilter::Consumed` returned nothing in restore.
+async fn submit_update_ger_note(
+    miden_client: &MidenClient,
+    accounts: crate::AccountsConfig,
+    ger_bytes: [u8; 32],
+) -> anyhow::Result<()> {
+    let inner_accounts = accounts.0.clone();
+    miden_client
+        .with(move |client| {
+            Box::new(async move {
+                client.sync_state().await?;
+                let ger_manager_id = inner_accounts
+                    .ger_manager
+                    .as_ref()
+                    .map(|a| a.0)
+                    .unwrap_or(inner_accounts.service.0);
+                let bridge_id = inner_accounts.bridge.0;
+                let ger = ExitRoot::new(ger_bytes);
+                let note = UpdateGerNote::create(ger, ger_manager_id, bridge_id, client.rng())?;
+                tracing::info!(
+                    note_id = %note.id(),
+                    ger = %hex::encode(ger_bytes),
+                    "UpdateGerNote created"
+                );
+                let tx_request = TransactionRequestBuilder::new()
+                    .own_output_notes(vec![note])
+                    .build()?;
+                let tx_id = client
+                    .submit_new_transaction(ger_manager_id, tx_request)
+                    .await?;
+                tracing::info!(
+                    tx_id = %tx_id,
+                    ger = %hex::encode(ger_bytes),
+                    "UpdateGerNote submitted, waiting for commit..."
+                );
+
+                let committed = crate::miden_client::wait_for_transaction_commit(
+                    client,
+                    tx_id,
+                    30,
+                    std::time::Duration::from_secs(1),
+                )
+                .await?;
+                if !committed {
+                    anyhow::bail!("UpdateGerNote tx {tx_id} not committed after 30s");
+                }
+                tracing::info!(tx_id = %tx_id, "UpdateGerNote transaction committed");
+                Ok(())
+            })
+        })
+        .await
+}
+
 #[allow(clippy::too_many_arguments)]
 pub async fn insert_ger(
     ger_bytes: [u8; 32],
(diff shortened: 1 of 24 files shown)


src/ger.rs
```diff
@@ -94,6 +94,73 @@ pub struct GerInsertResult {
...
+/// (separate from the service account that the NTX builder constantly
+/// mutates via claim processing) keeps the account state stable across
+/// GER submissions, so we don't need a fresh client per call.
+///
+/// Fresh-client-per-GER was removed because it shared the main sqlite
+/// and advanced the sync cursor past blocks where bridge NTX consumes
+/// the UpdateGerNote. The main client's subsequent sync_nullifiers only
+/// queries [current_cursor, tip], so those consumption events were never
+/// discovered and `NoteFilter::Consumed` returned nothing in restore.
+async fn submit_update_ger_note(
```
mandrigin, line 112: shouldn't we reimport here too in case of an error?

src/init.rs
```diff
@@ -126,8 +126,29 @@ async fn add_wallet(
...
+    // We use `Public` rather than `Network` because the latter is
+    // testnet/devnet-only on current upstream — local miden-node builds
+    // (and any production node not running with network-tx enabled) reject
+    // Network deployments with `Network transactions may not be submitted
+    // by users yet`. Public gives us the recovery property (state on-chain,
+    // import-by-id works) without the network-tx-builder semantics, which
+    // the proxy doesn't use anyway.
     let (auth_component, key_pair) = create_auth_component(client)?;
     let account = Account::builder(client.rng().draw_word().into())
+        .storage_mode(AccountStorageMode::Public)
```
mandrigin, line 151: how would this work in the already existing accounts? this will be ignored?

mandrigin (dismissed):
two questions, looks good otherwise

src/ger.rs
revitteth: Good question — the reimport+retry is in the caller `insert_ger` at [src/ger.rs:201-224](https://github.com/gateway-fm/miden-agglayer/blob/feat/v0.4.0-self-heal/src/ger.rs#L201-L224):

```rust
match submit_update_ger_note(miden_client, accounts.clone(), ger_bytes).await {
    Ok(()) => {}
    Err(err) if crate::account_recovery::is_recoverable_account_error(&err) => {
        // reimport ger_manager and retry the submission once
        crate::account_recovery::reimport_account(miden_client, ger_manager_id, "ger_manager").await?;
        submit_update_ger_note(miden_client, accounts.clone(), ger_bytes).await?;
    }
    Err(err) => return Err(err),
}
```

`submit_update_ger_note` is intentionally a pure inner operation (no retry logic) so the wrapper can run it twice — once eagerly, then again after `reimport_account` if the first attempt failed with a recoverable account-state error. The doc comment at lines 97-100 spells this out:

> Submit the actual UpdateGerNote Miden transaction. Factored out of `insert_ger` so the caller can run it twice — once eagerly, then again after `reimport_account` if the first attempt failed with a recoverable account-state error.

Empirically verified end-to-end in `scripts/e2e-account-self-heal.sh` — on `MODE=expect_self_heal`, the proxy logs show:
```
WARN ger.rs:204  GER injection: recoverable account error, reimporting ger_manager and retrying,
                err: account data wasn't found for account id 0x…
INFO account_recovery.rs:141  reimported from node, account: "ger_manager"
```
…and the deposit transitions to `ready_for_claim=true` on the retry.

src/init.rs
revitteth: Correct — `add_wallet` runs only at `--init` time (first deploy), so this change does **not** retroactively convert existing accounts. Bali's accounts were deployed pre-`dbe5c2d`-fix as `Private` and will stay `Private`; the runtime self-heal in this PR **cannot** recover them in place because `Client::import_account_by_id` returns `AccountIsPrivate` on Private accounts.

That's exactly why `docs/REDEPLOY_RUNBOOK_BALI.md` documents a **full redeploy** for bali rather than an in-place upgrade:

> v0.4.0's runtime self-heal CANNOT recover bali's current `ger_manager` because that account is Private (created pre-`dbe5c2d`-fix). Bali specifically needs a full redeploy with fresh Public accounts + aggkit reconfig.

Three-step procedure in the runbook:
1. Roll the proxy with `--init --reset-miden-store` (gets a clean Public-mode deployment).
2. Update aggkit config to point at the new `bridge_accounts.toml` account IDs.
3. Restart aggkit. Next aggoracle push hits the new proxy, GER injects cleanly, marti's deposit flips `ready_for_claim`.

For **future** deployments (and any other proxy redeployed off v0.4.0+), accounts are Public from the start — recoverable from sqlite loss, self-heal handles the bug class without operator intervention.

PR #76, opened 2026-06-09
fix(restore): decode GER limbs little-endian in restore_gers (Cantina MA#11)

revitteth:
## Cantina MA#11 — restore_gers rebuilds GER with wrong endianness

This held on `origin/main`: inside `restore_gers`, the X6-bounds-checked `Ok(v)` arm decoded each 4-byte GER limb with `v.to_be_bytes()` (`src/restore.rs:658` on `ea87178`). The prior X6 fix (`b076e14`) only added a `u32::try_from` bounds check; it never touched endianness, so the literal objection was still true.

### Root-cause / direction verification (done against pinned deps, not assumed)
- `UpdateGerNote::create` (miden-agglayer-0.14.4 `src/update_ger_note.rs:94`) stores `ExitRoot::to_elements()`.
- `ExitRoot::to_elements()` (miden-agglayer-0.14.4 `src/utils.rs:37`) routes through `miden_core::bytes_to_packed_u32_elements`.
- `bytes_to_packed_u32_elements` (miden-core-0.22.1 `src/utils/mod.rs:143`) packs each u32 with `u32::from_le_bytes` (documented "little-endian"); its documented inverse `packed_u32_elements_to_bytes` (line 172) uses `to_le_bytes`.

So the encode side is unambiguously little-endian, and the decode in `restore_gers` MUST mirror it with `to_le_bytes`. Big-endian decode byte-swapped every limb, rebuilding and republishing a GER hash that never existed on L1 — which can stall a fresh AggKit stack's injected-GER view after a `--restore` and freeze the bridge-in claim path.

### The diff
- `src/restore.rs:670`: `copy_from_slice(&v.to_be_bytes())` → `copy_from_slice(&v.to_le_bytes())` at the sole GER-rebuild limb-decode site. `to_be_bytes` is gone from the GER rebuild path. The X6 `u32::try_from` bounds check is retained. An explanatory MA#11 comment documents the encode/decode contract.

### The test
- `restore::tests::ma11_restore_ger_decoder_round_trips_little_endian` (`src/restore.rs:966`). Ported from the auditor PoC `restore_ger_decoder_byte_swaps_each_u32_limb_poc` (which asserted the BUGGY `be` behaviour) and inverted into the production assertion. It:
  - encodes via the real production encoder `ExitRoot::to_elements()` (identical to what `UpdateGerNote::create` writes),
  - decodes mirroring `restore_gers`,
  - asserts `restored == original` (passes ONLY with the LE fix),
  - regression guard: proves the old `to_be_bytes` decode byte-swaps each limb and yields a DIFFERENT root.
  - The test data has distinct per-limb bytes, so the round-trip assertion genuinely depends on the fix (it is not vacuously true).

### Build / test status
- Build: green. `Compiling miden-agglayer-service v0.1.0 ... Finished`.
- Test: `running 1 test ... test restore::tests::ma11_restore_ger_decoder_round_trips_little_endian ... ok` — `1 passed; 0 failed`.

### Verifier verdict
`addresses` / `makesObjectionFalse: true`. The verifier independently re-derived LE as the correct direction from the pinned deps, confirmed `to_be_bytes` is gone from the GER rebuild, and confirmed the test data depends on the fix. No `mustFix` items.

---

### DIFF
--- a/src/restore.rs
+++ b/src/restore.rs
@@ -655,7 +655,19 @@ async fn restore_gers(
                         // silently restoring the wrong root.
                         match u32::try_from(felt.as_canonical_u64()) {
                             Ok(v) => {
-                                ger_bytes[i * 4..(i + 1) * 4].copy_from_slice(&v.to_be_bytes())
+                                // MA#11 — the GER limbs are packed
+                                // little-endian when the note is created:
+                                // `UpdateGerNote::create` stores
+                                // `ExitRoot::to_elements()`, which is
+                                // `bytes_to_packed_u32_elements` using
+                                // `u32::from_le_bytes` (its documented inverse
+                                // `packed_u32_elements_to_bytes` uses
+                                // `to_le_bytes`). Decoding with `to_be_bytes`
+                                // byte-swapped every limb, rebuilding and
+                                // republishing a GER hash that never existed on
+                                // L1. Use `to_le_bytes` so restore round-trips
+                                // the original injected root.
+                                ger_bytes[i * 4..(i + 1) * 4].copy_from_slice(&v.to_le_bytes())
                             }
                             Err(_) => {
                                 tracing::error!(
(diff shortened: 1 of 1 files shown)


mandrigin:
Closing as **superseded by `main` (PR #94)**.

The identical fix is already on main: `src/restore.rs` `ger_bytes_from_storage` decodes GER limbs little-endian — `ger_bytes[i*4..(i+1)*4].copy_from_slice(&v.to_le_bytes())` (commit `2b193e9`, merged in the synthetic-indexer redesign). The helper carries a doc-comment explaining the MA#11 byte-swap bug and is unit-tested via a round-trip against `ExitRoot::to_elements`.

This PR also edits the **old inline-loop location** (`match u32::try_from` around the former line 658) that no longer exists on main — it was refactored into `ger_bytes_from_storage` — so it would conflict. The regression test is further covered by open PR #109 (`finding_11_ger_restore_roundtrip_le_not_be`, a non-symmetric GER so LE≠BE in every limb). Closing.

PR #79, opened 2026-06-09
fix(store): store-owned atomic synthetic block allocation (Cantina MA#5/#19/#15)

revitteth:
## Cantina MA#5 / MA#19 / MA#15 — store-owned atomic synthetic block allocation

Cluster A. cergyk rejected the previous fixes as insufficient (and called MA#19 a probable "AI hallucination"). These three findings share one architectural root: synthetic EVM block numbers were reserved with a non-atomic `get_latest_block_number() + 1` **outside** the store transaction, then used by multiple writers. This PR makes block **allocation itself** store-owned and transactional, hoists the per-note increment out of the bridge-out loop, and folds the `deposit_count` increment into the same transaction as the log write (idempotent on retry).

Single commit: `1ecbd3d` — `fix(store): store-owned atomic synthetic block allocation (Cantina MA#5/#19/#15)`.

---

### MA#5 (HIGH) — Non-atomic synthetic block allocation can hide committed bridge-outs

**cergyk / Alex Keller objection:** *"block numbers are still chosen outside the store transaction with `get_latest_block_number()+1`. Two writers can still pick the same next block ... insert a late BridgeEvent into a block a downstream poller may already have processed ... A complete fix should make block allocation itself store-owned and transactional."*

**Diff:** All four synthetic writers (`bridge_out.rs`, `ger.rs`, `claim.rs`, `claim_watcher.rs`) no longer compute `get_latest_block_number()+1` for the event block. The three atomic commit helpers (`commit_b2agg_event_atomic`, `commit_ger_event_atomic`, `commit_manual_claim_event_atomic`) allocate the synthetic block **inside** the same transaction/critical section as the log write, via the store-owned `advance_block_number` / `allocate_synthetic_block`. InMemoryStore holds the `latest_block_number` write lock across the increment; PgStore uses `UPDATE service_state SET latest_block_number = latest_block_number + 1 ... RETURNING` inside the SERIALIZABLE commit transaction — so two concurrent writers cannot observe the same tip and both pick N+1. Raw Miden sync height is now tracked in a **separate** `raw_miden_height` column (migration `008`, additive/backward-compatible); `StoreSyncListener` advances the synthetic tip monotonically rather than clobbering it with the raw Miden number.

**Test:** `store::memory::tests::cantina_5_concurrent_allocations_get_distinct_blocks` (positive/post-fix form of auditor `poc-05`) — two writers reserving from the same observed tip get DISTINCT blocks (101 then 102), 64 further reservations are all unique, and a late `BridgeEvent` lands in its own block 102 with a distinct hash, never the already-visible block 101. Plus `cantina_5_raw_miden_height_is_separate_from_synthetic_tip`.

**Status:** build green; `cargo test --lib cantina_5` => 5 passed, 0 failed.

---

### MA#19 (MEDIUM) — In-loop block increment for bridge out (authored by cergyk)

**cergyk objection:** *"This is trivially not fixed, probably due to an AI hallucination. Please read the recommendation linked in this report."* The recommendation was to hoist the block increment out of the per-note loop.

**Diff:** The offending pair in `bridge_out.rs::on_post_sync` — `let block_number = self.store.get_latest_block_number().await? + 1; ... if self.process_consumed_note(note, block_number).await { self.store.set_latest_block_number(block_number).await?; }` **inside** `for note in &consumed_notes` — is deleted. The loop now shares a single, lazily-allocated `batch_block: Option<u64>` across the whole batch; the tip advances by exactly one per tick regardless of note count (a 1000-note Miden tx advances by 1, not 1000). Lazy allocation preserves the no-phantom-block invariant on all-skip ticks.

**Test:** covered by the shared-block / distinct-block assertions in `cantina_5_concurrent_allocations_get_distinct_blocks`; the production assertion is that a batch lands in one block and the tip does not jump by note count.

**Status:** build green; verified by the cantina_5 suite above.

---

### MA#15 (LOW) — Non-atomic processed-note bookkeeping can renumber committed bridge-outs

**cergyk objection:** *"The deposit_count increment is not included in the add_log transaction, thus the updates are not atomic."*

**Diff:** `PgStore::commit_b2agg_event_atomic` now folds the `deposit_count` allocation (`mark_note_processed`) and the bridge-event log insert into ONE postgres transaction (`txn.commit()` spans both), and is idempotent: `get_processed_deposit_count` is checked first, reusing the assigned count and emitting no duplicate log on retry. `mark_note_processed` switched from a `HashSet` to a `note_id -> deposit_count` `HashMap` so the assigned count is reusable. InMemoryStore mirrors this. `restore_bridge_outs` uses the same helper.

**Test:** `store::memory::tests::cantina_15_b2agg_commit_is_atomic_and_idempotent_on_retry` — first commit assigns `deposit_count` 0; retry of the same note REUSES 0, does NOT re-increment the counter, emits NO duplicate log; a distinct note gets the next count (1).

**Status:** build green; `cargo test --lib cantina_15_b2agg_commit_is_atomic_and_idempotent_on_retry` => 1 passed.

---

### Verification (run independently, not just trusting the implementer report)

- `cargo build` (this worktree) => Finished, 0 errors.
- `cargo test --lib cantina_5 -- --nocapture` => **5 passed, 0 failed** (incl. the two new store tests).
- `cargo test --lib cantina_15_b2agg_commit_is_atomic_and_idempotent_on_retry -- --nocapture` => **1 passed**.
- `git grep "get_latest_block_number().await? + 1"` over `src` => **no matches** (no surviving non-atomic synthetic-allocation pattern).

### Caveats (honest, not blockers)

- The PgStore atomic-commit + `raw_miden_height` SQL paths were verified by reading the SQL (no live `DATABASE_URL` run in this environment). The InMemoryStore analogue that actually executes mirrors the same logic and passes.
- `restore.rs` numbering change: claims and GERs now each get their own store-allocated block instead of sharing one fixed `restore_block`. This is a behavioral change that matches the live path; non-regressive for #5/#19/#15.

### Outstanding (mustFix)

None.

### DIFF
--- a/migrations/008_raw_miden_height.sql
+++ b/migrations/008_raw_miden_height.sql
@@ -1,0 +1,12 @@
+-- Cantina #5 — track the raw Miden sync height SEPARATELY from the synthetic
+-- EVM tip (service_state.latest_block_number).
+--
+-- Pre-fix the sync loop (StoreSyncListener::on_post_sync) wrote the raw Miden
+-- block number directly into latest_block_number, conflating the two counters
+-- and risking the synthetic tip being rolled backwards below a block a
+-- store-owned allocator had already published a synthetic log into. The
+-- synthetic tip is now owned by the atomic commit helpers (which allocate the
+-- next synthetic block inside their transaction); the raw Miden height lives
+-- here and is only ever observed, never used to clobber the synthetic tip.
+ALTER TABLE service_state
+    ADD COLUMN IF NOT EXISTS raw_miden_height BIGINT NOT NULL DEFAULT 0;
--- a/src/block_state.rs
+++ b/src/block_state.rs
@@ -164,6 +164,14 @@ impl BlockState {
         GENESIS_TIMESTAMP + block_num * BLOCK_TIME
     }
 
+    /// Pure synthetic block timestamp for a given block number. Exposed as an
+    /// associated function so the store can derive the timestamp from a
+    /// store-allocated synthetic block number without holding a `BlockState`
+    /// (Cantina #5 — synthetic block allocation is store-owned).
+    pub fn synthetic_timestamp(block_num: u64) -> u64 {
+        Self::deterministic_timestamp(block_num)
+    }
+
     /// Compute the deterministic timestamp for any block number.
     pub fn get_block_timestamp(&self, block_num: u64) -> u64 {
         Self::deterministic_timestamp(block_num)
--- a/src/bridge_out.rs
+++ b/src/bridge_out.rs
@@ -266,18 +266,30 @@ impl BridgeOutScanner {
         destination_network == self.local_network_id
     }
 
-    /// Process a consumed B2AGG note. Returns `true` if the caller should advance
-    /// `latest_block_number` (a synthetic log was written at `block_number`),
-    /// `false` if the caller must NOT advance (the note was skipped, was a
-    /// non-B2AGG, errored on parse, or was a self-target poison leaf).
+    /// Process a consumed B2AGG note into the store-allocated batch block.
+    /// Returns `true` if a synthetic `BridgeEvent` log was written, `false` if
+    /// the note was skipped (non-B2AGG, parse error, reclaim, self-target
+    /// poison leaf, unknown faucet, etc).
     ///
-    /// Self-review: pre-fix this returned `()`, and the caller in `on_post_sync`
-    /// unconditionally bumped `latest_block_number` afterwards. When the
-    /// self-target circuit-break (Cantina #13) added an early return without a
-    /// log write, every poison leaf left a phantom block: `eth_blockNumber`
-    /// advanced but no log existed at that block, breaking the "every reader
-    /// who sees latest >= N also sees the log at N" invariant.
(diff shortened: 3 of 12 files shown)


mandrigin:
Closing as **superseded by `main` (PR #94, synthetic-indexer redesign)**.

This PR's entire premise — multiple synthetic writers reserving blocks via non-atomic `get_latest_block_number() + 1` — no longer exists on main. `git grep "get_latest_block_number().await? + 1"` on `main` returns nothing. `SyntheticProjector` is now the **sole ordered producer**: it sets `set_latest_block_number(miden_block)` with block N == Miden height N (Miden-1:1), with no `get_latest()+1` reservation race by construction. That solves MA#5 and MA#19 (the in-loop increment path was gutted along with the old `bridge_out.rs` emit loop).

The machinery this PR builds (store-owned `advance_block_number`, `commit_b2agg_event_atomic`, `ClaimWatcher`) is vestigial on main and the PR rewrites emission paths the redesign removed (CONFLICTING). **Do not rebase** — it would re-introduce a superseded allocation architecture.

Residual: if we want MA#15's strict single-transaction `deposit_count`+log semantics, that's a small standalone follow-up, not this PR. Closing.

PR #98, opened 2026-06-30
Fix/miden store db lock

mandrigin:
## What / why

`RD-1112` — the miden-client sqlite store wedges on `database is locked` under the proxy's concurrent access. Confirmed **root cause of PRST-4035**: prod telemetry (outpost-testnet, 2026-06-30) showed `MidenClient::sync` stalled ~15 min on `SQLITE_BUSY` (19 hits, backoff pinned at the 60s cap), so sync never reached the bridge-out block → the exit went unobserved.

Root cause is upstream: `miden-client-sqlite-store`'s `pool_manager::new_connection` opens every pooled connection with only `PRAGMA foreign_keys=ON` — no `journal_mode` (→ rollback-journal, where a reader's SHARED lock blocks a writer's COMMIT), no `busy_timeout` override, and no config knob. The proxy reads the store from its Miden sync thread while the RPC claim path commits a write → the commit is blocked into `SQLITE_BUSY`.

## What's in this PR

- **RED reproduction** (`src/sqlite_pragmas.rs`): `writer_commit_not_blocked_by_concurrent_reader` (raw rusqlite) and `real_miden_store_write_locked_by_concurrent_reader` (drives the **actual** `miden-client-sqlite-store::SqliteStore` into the prod `database is locked` in ~5s). Both go GREEN once the store is opened in WAL.
- **RD-1112 observability** (`fe08a04`): `trace_store_locked()` emits a stable `RD-1112:` log prefix + bumps `miden_store_locked_total{path}`; no-op for non-lock errors. Wired into both catch sites — `MidenClient::sync` (the path seen in prod) and `wait_for_transaction_commit` (the claim-path post-submit sync, which previously propagated the lock with **no** log). Makes it a single Grafana/Loki query + rate-alertable metric, and stays as a regression tripwire after the fix.
- **Vendored loadtest + handoff** (`DB-LOCK-WIP.md`, `scripts/e2e-bridge-loadtest.sh`): N=250 concurrent-bridge loadtest and the continue-anywhere runbook.

## Pending (not in this PR yet)

- **GREEN — the WAL stopgap**: open the store once in `journal_mode=WAL` at `MidenClient::new` startup (persists in the DB header → the whole pool inherits it). Flips both tests RED→GREEN.
- **Upstream PR** to `0xMiden/miden-client-sqlite-store` adding `journal_mode=WAL` (+ real `busy_timeout`) to `new_connection`.

No behavior change from the observability commit. Base: `release/0.15`.

### DIFF
--- a/DB-LOCK-WIP.md
+++ b/DB-LOCK-WIP.md
@@ -1,0 +1,32 @@
+# DB-lock WIP — handoff (`fix/miden-store-db-lock`, off `v0.15.1`)
+
+**Status:** RED reproduced (unit); RED e2e in progress; GREEN (WAL fix) **not yet applied**.
+
+## The bug
+The proxy opens the miden-client sqlite store (`<store-dir>/store.sqlite3`) from **two of its own contexts** — the Miden **sync thread** (`MidenClient` loop, `sync_state()`) and the **RPC claim path** (`src/service_send_raw_txn.rs`). The store runs in default **rollback-journal** mode: `miden-client-sqlite-store`'s pool manager sets only `foreign_keys=ON`, and rusqlite's default 5s `busy_timeout` applies. In rollback-journal mode a reader's SHARED lock blocks a writer's COMMIT (which needs EXCLUSIVE), so under concurrent load the claim-path COMMIT is blocked by the sync reader and eventually returns `SQLITE_BUSY` → `database is locked`. No external process is involved.
+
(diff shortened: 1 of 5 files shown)


mandrigin:
Superseded by #94. The db-lock fix shipped there as a true-singleton `MidenClient` (serialized actor), tested and confirmed good under load — WAL was not needed. RD-1112 remains the tracking ticket; the observability tripwire (log tag + `miden_store_locked_total` counter) lives on the debug branch if it's ever wanted as a regression canary.

PR #129, opened 2026-07-09
PoC(Cantina #7): pre-emit LET assignment gate — never emit a misnumbered depositCount

mandrigin:
## PoC — Cantina #7: LET assignment gate (alarm → pre-emit gate)

**Draft / proof-of-concept.** Finding #7's failure mode is a `depositCount` that diverges from the bridge's Local Exit Tree: the index is *inferred* from consumed-note enumeration, and once one on-chain leaf is missed, the next exit is misnumbered and **every exit after it shifts** — unclaimable withdrawals, silently. The Cantina #9 monitor already compares local vs. on-chain, but only **after** emission (alarm-only).

This PoC promotes that comparison to a **pre-emit assignment gate** in the projector: no `depositCount` is ever handed out while the local view and the on-chain LET disagree. A stall is recoverable (reconciler / MA#18 recovery closes the gap, next tick proceeds); a wrongly-numbered BridgeEvent is poison.

### The invariant
Only the bridge's consumption of a B2AGG note appends a LET leaf (reclaims never touch bridge storage). Emitted notes advanced `deposit_counter`; visible-but-pending ones (this tick's batch, quarantined, deferred) will. So alignment collapses to one comparison, with **no extra RPC**:

```
count(visible bridge-consumed B2AGG notes)  ==  AggLayerBridge::read_let_num_leaves(bridge)
```

both sides read from the **same synced client-store snapshot** the tick already fetched — no assignment-vs-feed race. `direct_recovered` (spent-before-import) notes are counted in explicitly since they live outside the client store.

### Behavior
- **Aligned** → project normally (zero overhead beyond one local account read + one pass over the already-fetched feed).
- **InvisibleGap** (on-chain leaves with no visible note — erased/undelivered exits) → **halt the tick before any emission**, `bridge_let_assignment_gate_halted_total{kind="invisible_gap"}` + error log; cursor unchanged, retried next tick.
- **LocalAhead** (more visible consumptions than leaves — store corruption / misattribution) → fail closed, same halt with `kind="local_ahead"`.
- Bridge account not yet tracked (first boot) → gate skips (nothing can have been consumed).

### Why this refutes the "requires upstream API" claim
The gate uses **only existing surface**: `client.get_account` + `AggLayerBridge::read_let_num_leaves` (miden-agglayer 0.15.3) — the same call the #9 monitor has used in production since #94. No upstream change needed for the live path.

### Changes
- `src/synthetic_projector.rs`: `LetGateVerdict` + pure `let_assignment_gate(visible, on_chain)` predicate; gate wired into `tick` after the feed/buckets are built, before the projection loop.
- `src/metrics.rs`: `bridge_let_assignment_gate_halted_total` registered.
- Unit test `cantina7_let_assignment_gate_verdicts` (aligned incl. genesis, gap, local-ahead).

### PoC caveats (deliberate scope cuts)
- A halt stalls the **whole** tick (claims + GERs too, not just B2AGG) — conservative; a production version could park only B2AGG emission.
- Quarantined/deferred notes keep the equation balanced (they're visible), so the gate does **not** fire on the known-tolerated divergences — only on *invisible* leaves, which is exactly finding #7's trigger.
- Restore-time per-block reconstruction (walking `GetBlockByNumber` account deltas) is out of scope here; the gate covers the live path.
- Gated by `make lint` only so far (fmt + clippy `-D warnings` clean; unit test green). No e2e yet.

Refs: Cantina finding #7 (Acknowledged), Cantina #9 monitor (`run_let_divergence_check`), #119 (H4 ordering), MA#18 recovery (#126).

---

## Update — same-tx B2AGG ordering (commit 4652bf2)

Addresses the one residual the cardinality gate can't see: **two B2AGG notes consumed by the *same* bridge transaction**. Their `consumed_tx_order` ties (it's per-tx), so the sort previously fell back to the details-commitment (a hash) — the two `depositCount`s could be **swapped**, both exits unclaimable.

Fix uses existing surface only: the bridge's per-account tx feed carries the order — each header's `input_notes()` list is the on-chain LET append order.
- `bridge_consumed_nullifiers` now returns `(block, tx_order, within_tx_pos)`.
- `tick` fetches `sync_transactions` **only when a real same-tx tie exists** (≥2 emit-class B2AGG with equal `(block, tx_order)`); otherwise zero extra RPC. A tied note missing from the feed → **halt** (`kind="unordered_tie"`), same fail-closed posture as the gate.
- sort key is now `(consumed_tx_order, within_tx_pos, details_commitment)`.

**Documented PoC limitation:** a reconciler-fabricated `ConsumedExternalNoteState` record carries no metadata, so `InputNoteRecord::nullifier()` returns `None` and it can't be matched to the tx feed — it keeps the commitment tie-break. A production version would additionally map `nullifier → details_commitment` (derivable per tx-input note) so metadata-less records also order exactly.

### Coverage matrix (now)
| Scenario | depositCount correct? | By |
|---|---|---|
| 2 B2AGG, different txs, same block | ✅ | `consumed_tx_order` |
| 2 B2AGG, same tx | ✅ (with metadata) / commitment fallback (metadata-less — see limitation) | within-tx position |
| Invisible / late consumption | ✅ halt, never misnumber | cardinality gate |
| Gap/shift after a missed leaf | ✅ halt | cardinality gate |

### DIFF
--- a/src/metrics.rs
+++ b/src/metrics.rs
@@ -93,6 +93,15 @@ pub fn init_metrics() {
          kind=aggkit_ahead (local state corruption)."
     );
     describe_counter!(
+        "bridge_let_assignment_gate_halted_total",
+        "Projector ticks halted by the pre-emit LET assignment gate (Cantina #7 \
+         PoC): depositCount assignment refused because the on-chain Local Exit \
+         Tree and the visible bridge-consumed B2AGG set disagree. Labels: \
+         kind=invisible_gap (on-chain leaves with no visible note — erased or \
+         undelivered exits) or kind=local_ahead (more visible consumptions than \
+         leaves — store corruption; fail closed)."
+    );
+    describe_counter!(
         "bridge_burn_serial_collision_total",
         "BURN note serial collisions (Cantina #5). Each increment marks \
          a BURN note whose serial number was already observed for a \
--- a/src/synthetic_projector.rs
+++ b/src/synthetic_projector.rs
@@ -47,7 +47,7 @@
 use crate::accounts_config::AccountsConfig;
 use crate::block_state::BlockState;
 use crate::bridge_address::get_bridge_address;
-use crate::bridge_out::is_b2agg_note;
+use crate::bridge_out::{B2AggConsumerClass, classify_b2agg_consumer, is_b2agg_note};
 use crate::miden_client::{MidenClientLib, SyncListener};
 use crate::restore::{
     B2AggRestoreOutcome, ClaimProjectOutcome, GerProjectOutcome, project_b2agg_note,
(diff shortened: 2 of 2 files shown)


mandrigin:
Closing this PoC as superseded by #137, which carries the production-oriented Cantina #7 implementation and its follow-up hardening. Keeping the branch/history intact for reference.

PR #130, opened 2026-07-10
fix(projector): defer late-consumption sweep while caught up

mandrigin:
## Summary

- Defer late-consumption projection when the projector cursor is already at the chain tip.
- Do not mark deferred note IDs as swept before a block-advancing tick can project them.
- Retry automatically on the next advancing tick, keeping the delay bounded by the block cadence.

## Problem

Late note consumptions can enter the client store after the projector cursor has already passed their consumed block. The recovery sweep buckets those notes at cursor + 1, but that bucket is only processed while the cursor advances.

When a tick runs with cursor equal to tip, the projection loop does not execute. Previously, the note IDs were still added to the in-memory swept set, permanently suppressing their BridgeEvents until the process restarted and cleared that cache.

## Fix

When late notes are found while the projector is caught up, leave them unbucketed and unswept. The next block-advancing tick discovers them again, projects the cursor + 1 bucket, and only then records their IDs as swept.

## Validation

Live forensics reproduced the failure under concurrent load: the note was committed, consumed, and visible on-chain, while the synthetic BridgeEvent and deposit counter lagged by one. Restarting the proxy cleared the swept cache and restored both within 11 seconds, confirming the drop mechanism this change removes.

## Scope

One projector code path in src/synthetic_projector.rs; no API, schema, or storage-format changes.

### DIFF
--- a/src/synthetic_projector.rs
+++ b/src/synthetic_projector.rs
@@ -1115,19 +1115,42 @@ impl SyntheticProjector {
                 })
                 .filter(|n| !swept.contains(&n.details_commitment().as_bytes()))
                 .collect();
-            let ids = late
-                .iter()
-                .map(|n| n.details_commitment().as_bytes())
-                .collect();
-            if !late.is_empty() {
+            // #27 (2026-07-09): the late bucket lands at `cursor + 1`, which only
+            // the `while cursor < tip` loop below projects. When the projector is
+            // CAUGHT UP (`cursor == tip` — true on most ticks, since ticks are ~1s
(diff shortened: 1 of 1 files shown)


mandrigin:
Closing as superseded by #134, the canonical release-stack PR. This PR's sole late-sweep patch is reproduced exactly in #134 (`7258a6a` and `353df6f` have the same stable patch ID), so no implementation is being dropped. Historical discussion remains here; further review and fixes should continue on #134.

PR #132, opened 2026-07-10
fix(projector): gate synthetic block sealing on reconciliation frontier

mandrigin:
## Summary

Gate synthetic block sealing on the note reconciler's completed frontier.

When node RPC reconciliation is enabled, the projector now advances only through:

```
min(Miden sync tip, reconcile cursor)
```

This prevents future synthetic blocks from being sealed before the public tag-0 note sweep for those blocks has completed. The change also adds reconciliation-backlog and late-sweep observability metrics.

## Problem this helps with

An externally created public B2AGG note can be committed and consumed between local client sync points. The reconciler can recover that note later, but if the projector has already sealed its Miden block, the resulting BridgeEvent must be emitted in a later synthetic block.

That preserves eventual delivery but breaks the Miden-1:1 block-placement contract used by forward-only log consumers. Holding the synthetic frontier converts that case into visible head lag: reconciliation completes first, then the original block is projected with the note available.

## Operational tradeoff

This is deliberately safety over liveness:

- normal caught-up operation is unchanged
- reconciliation lag or errors hold eth_blockNumber and synthetic-log visibility
- projection throughput is bounded by reconciliation throughput
- the held-block gauge surfaces that backlog

## Scope and non-goals

- Forward-looking only: it cannot repair events in blocks already sealed before the barrier became active.
- Requires node RPC reconciliation; node_rpc=None keeps legacy behavior.
- Covers notes observable through the reconciler's public tag-0 scan; it does not make erased or otherwise node-invisible notes visible.
- Does not add multi-replica projector safety.
- CLAIM and GER notes were already locally visible; the primary benefit is external B2AGG visibility and exact placement.

## Known review gaps

This is a draft because the early review found issues that must be resolved before the guarantee is trustworthy:

- the persisted reconcile cursor can advance past a spent-before-import recovery held only in memory; a crash in that window can lose the queued event
- omitting --miden-node is a valid localhost configuration for MidenClient but currently gives the projector node_rpc=None, silently bypassing the barrier
- receipts are still finalized at the raw Miden sync height, so a held synthetic head can expose a receipt for a block above eth_blockNumber
- projector_late_sweep_anomaly_total currently classifies normal already-projected notes as anomalies, especially after restart
- a persisted reconcile cursor ahead of the current node tip is treated as caught up and fails open
- persistent reconciliation failure becomes an indefinite projection freeze and needs an explicit readiness/alerting contract
- no barrier-specific regression tests are included yet

## Dependency

Stacked on #130 / fix/late-sweep-caught-up-drop so this PR contains only the visibility-barrier commit. PR #130 remains the late-sweep fail-safe path and is still under review.

## Validation

- cargo fmt --all -- --check: pass
- git diff --check: pass
- inherited synthetic-projector tests: 14/14 pass
- inherited metrics tests: 14/14 pass
- the inherited tests do not exercise the new tick barrier path; no live barrier/fault-injection validation is claimed

### DIFF
--- a/src/metrics.rs
+++ b/src/metrics.rs
@@ -63,6 +63,16 @@ pub fn init_metrics() {
     describe_counter!("rpc_requests_total", "Total JSON-RPC requests by method");
     describe_counter!("claims_processed_total", "Total claims processed");
     describe_counter!("ger_injections_total", "Total GER injections");
+    describe_gauge!(
+        "projector_visibility_barrier_held_blocks",
+        "#30 visibility barrier: blocks the projector is holding because the reconciler \
+         sweep cursor is behind the Miden tip (0 in steady state; >0 signals reconciler lag)"
+    );
+    describe_counter!(
+        "projector_late_sweep_anomaly_total",
+        "#30: notes the late-consumption sweep found for an already-sealed block — the \
+         visibility barrier should make this impossible, so any increment is a bug signal"
+    );
     describe_counter!("bridge_outs_total", "Total bridge-out operations");
     describe_counter!("store_errors_total", "Total store operation errors");
     describe_histogram!("rpc_request_duration_seconds", "JSON-RPC request duration");
--- a/src/synthetic_projector.rs
+++ b/src/synthetic_projector.rs
@@ -1055,7 +1055,41 @@ impl SyntheticProjector {
                 "note reconciler failed (transient — will retry next tick)"
             );
         }
-        if cursor >= tip {
+        // #30 VISIBILITY BARRIER: never seal a synthetic block the reconciler
+        // has not fully swept. `reconcile_notes` (above) advances
+        // `reconcile_cursor` to its last completed sweep window; every external
+        // bridge-out note at a block <= reconcile_cursor is therefore already in
+        // the client store. Capping the projection loop at `min(tip,
+        // reconcile_cursor)` makes exact-block emission a GUARANTEE for all event
+        // types: a note is always visible BEFORE its block is projected, so
+        // nothing is ever "late" (proxy-created CLAIM/GER notes are instant-
+        // visible and were never late; B2AGG lateness was purely import lag,
+        // eliminated here by construction — the late sweep below becomes an
+        // alarm). In steady state the reconciler reaches `tip` every tick
+        // (sub-second ~1000-block windows), so the barrier is a no-op; it only
+        // bites when the reconciler falls behind, and then holding is the
+        // correct failure mode — a late synthetic tip is benign (aggkit reads
+        // block ranges and just sees the chain pause), an event at the wrong
+        // block is poison. No reconciler wired (node_rpc = None, pure unit
+        // tests / a non-reconciling deployment) => no barrier, legacy behavior.
+        let project_to = if self.node_rpc.is_some() {
+            let reconcile_cursor = self.reconcile_cursor.load(Ordering::Acquire);
+            let held = tip.saturating_sub(reconcile_cursor);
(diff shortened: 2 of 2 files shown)


mandrigin (commented):
## Early visibility-barrier review

The low-water-mark idea is directionally right: for future blocks, projecting only through min(Miden tip, reconcile cursor) can prevent ordinary public tag-0/B2AGG notes missed by local sync from being discovered only after their synthetic block was sealed. Converting that race into explicit synthetic-head lag is a sensible safety-over-liveness trade.

I do not think the current branch is ready to merge, though:

- [ ] **Make direct recovery crash-safe.** The reconciler can reconstruct a spent-before-import B2AGG into the in-memory-only direct_recovered queue, then persist/advance reconcile_cursor before tick projects that queue. A crash after frontier persistence but before projection loses the volatile record; restart trusts the advanced frontier and can seal the block without its BridgeEvent. Persist pending recoveries, or advance the durable reconcile frontier only after recovery work through that frontier is durably projected. Add a crash/restart regression test.

- [ ] **Enable the barrier for the documented default node endpoint.** The CLI allows miden_node=None and MidenClient correctly interprets that as localhost, but SyntheticProjector maps the same None to node_rpc=None. The new branch then explicitly uses legacy project_to=tip, silently disabling both reconciliation and the barrier in a valid default deployment. Resolve the endpoint once and give both clients the same concrete node configuration.

- [ ] **Keep Ethereum RPC views coherent while the head is held.** StoreSyncListener still finalizes/expires receipts at the raw Miden sync block before the projector runs. During reconciliation lag this can expose a receipt in block N while eth_blockNumber is below N and eth_getBlockByNumber(N) returns null; BlockState may still make the same block available by hash. Receipt/block visibility should follow the sealed frontier, or this API deviation needs an explicit design and regression tests.

- [ ] **Fix the anomaly metric before treating it as an invariant alarm.** The late set includes every consumed note at or below cursor that is not yet in the in-memory swept cache. On-time notes are not inserted into swept when first projected, so they are classified and counted as anomalies on a later advancing tick before store dedup rejects the replay. After restart, swept is empty and historical consumed notes are counted again. Count only notes proven unprocessed/late, or persist/hydrate the relevant seen state.

- [ ] **Add barrier-specific tests.** This commit changes production code only. Existing test_projector instances use node_rpc=None and no test drives tick through the new barrier branch. At minimum cover: frontier behind/equal/ahead of tip; failed reconciliation window holding projection at the last good frontier; default localhost configuration; restart with pending direct recovery; receipt/head consistency; and normal on-time notes not incrementing the anomaly counter.

Additional scope/operational points:

- This is prospective, not retroactive: it cannot repair blocks already sealed before the frontier was established.
- It covers notes visible to the public tag-0 sweep, not erased or otherwise node-invisible notes.
- A persistent reconciler error now becomes an indefinite synthetic-chain freeze while on_post_sync still returns success. Sustained held-block lag should fail readiness and page operators.
- If a persisted reconcile cursor belongs to a different/ahead chain tip, min(tip, cursor) fails open; a chain/reset fingerprint or explicit reset invariant is needed.
- The caught-up deferral in parent PR #130 is still before an early return and remains unreachable. This barrier attacks the root ordering problem prospectively, but does not make that fallback execute.

Once the durable frontier and RPC-visibility issues are fixed, this should materially improve exact placement for external B2AGG events. Until then, the absolute “exact-block guarantee for all event types” claim is too broad.

mandrigin:
Closing as superseded by #134, the canonical release-stack PR. The visibility-barrier patch is reproduced exactly in #134 (`12c3241` and `2408441` have the same stable patch ID), and #134 also carries its late-sweep prerequisite on the current #133 stack. No implementation is being dropped; further review and fixes should continue on #134.

PR #145, opened 2026-07-19
fix(release): harden recovery, projection, and chaos follow-ups

mandrigin:
## Context

Post-`v0.15.8` hardening from failures observed on the real long-running multi-chain stack and during chaos/recovery testing. This is not speculative cleanup: it closes concrete recovery wedges, projection gaps, stale-nonce behavior, and false confidence in the soak harness.

## Runtime fixes

| Finding | Change | Operational effect |
|---|---|---|
| #67 | Fail-close CLAIM projection until full `claimAsset` calldata is recoverable; repair historical real-hash envelopes even at the projector tip | Prevents immutable ClaimEvents with unusable calldata from wedging AggKit recovery |
| #66 | Check the reserved-but-unemitted LET frontier immediately before each synthetic block seals | Prevents within-tick `depositCount` gaps and lets crash-reserved leaves heal by re-projection |
| #62 | Select metadata-recovery RPC by token `origin_network` | Restores L2B-origin ERC-20 metadata from L2B instead of incorrectly querying L1 |
| native recovery | Accept both AggLayer wrapped faucets and native Miden `FungibleFaucet`s as hash-gated metadata sources | Allows native Miden-originated assets to survive full restore |
| #69 | Recover historical bridge-consumed CLAIMs from the node scan and replay them in original Miden execution order | Rebuilds ClaimEvents after `--reset-miden-store`, allowing aggsender certificate recovery to resume |
| #55 | Reclaim an expired nonce reservation only when its original hash has no durable transaction row | Repairs the observed crashed-before-admission external-submitter wedge |

## Recovery and chaos validation

- replaces preservation of proxy nonce tables with a realistic bridge-service database resync after full proxy recovery;
- runs L2↔L2 and native-token tiers before the from-scratch recovery so the rebuild covers L1-, L2B-, and Miden-originated assets;
- retries garbo injections through proxy/PostgreSQL faults and requires enabled injection classes to actually fire;
- extends L2↔L2 settlement waits only while certificate/ClaimEvent progress remains observable;
- corroborates the external completeness verifier with proxy-store state and surfaces a silent `ntx-builder` as an explicit warning;
- fails the chaos run early when its completeness tool is unavailable.

```mermaid
flowchart LR
    Node[Miden node history] --> Restore[Full restore]
    L1[L1 token RPC] --> Metadata[Origin-aware metadata recovery]
    L2B[L2B token RPC] --> Metadata
    Metadata --> Restore
    Restore --> Claims[Historical ClaimEvents + calldata]
    Restore --> Exits[BridgeEvents]

    Consumed[Consumed Miden block] --> Projector[Synthetic projector]
    Projector --> Gate{All reserved LET leaves emitted?}
    Gate -->|yes| Seal[Seal synthetic block]
    Gate -->|no| Halt[Fail closed and retry]
    Seal --> AggKit[AggKit / aggsender]

    Chaos[Fault + garbo traffic] --> Verdict[Verifier + store + liveness verdict]
```

## Evidence captured by the branch

- full recovery on the grown ~25.7k-block stack restored 326 historical claims and allowed aggsender initial-state recovery to complete;
- L2B-origin and native Miden-originated token metadata were recovered byte-for-byte after a full database/store reset;
- post-recovery bridge traffic resumed in both directions;
- focused recovery, projector, nonce-reservation, metadata, PostgreSQL-gated, and shell-harness regressions are included.

This PR is initially a draft while the combined release-follow-up diff and CI are reviewed.

### DIFF
--- a/scripts/e2e-chaos-soak.sh
+++ b/scripts/e2e-chaos-soak.sh
@@ -39,6 +39,10 @@ POST_CHAOS_SETTLE="${POST_CHAOS_SETTLE:-150}"
 L2L2_FWD="${L2L2_FWD:-2}"
 L2L2_BACK="${L2L2_BACK:-2}"
 FRESH="${FRESH:-0}"
+# PR#145: PASS certifies EXACT-BLOCK fidelity by default — late events fail.
+# A consciously-degraded run (e.g. a grown/recovered stack) may set ALLOW_LATE=1;
+# the verdict line prints the mode either way.
+ALLOW_LATE="${ALLOW_LATE:-0}"
 TOOL_BIN="${TOOL_BIN:-$PROJECT_DIR/target/debug/bridge-out-tool}"   # repo-local default; override with $TOOL_BIN
 # #41: fail FAST if the debug tool is missing — a late WARN used to let the whole
 # storm run and then skip the completeness verdict entirely.
@@ -95,7 +99,7 @@ GARBO_PID=$!
 # split the soak's N evenly across L1->Miden / Miden->L1.
 say "=== mixed loadtest under storm (L1 ${N} split $((N / 2))/$((N - N / 2)), L2<->L2 $L2L2_FWD/$L2L2_BACK) ==="
 N_L1_FWD=$((N / 2)) N_L1_BACK=$((N - N / 2)) L2L2_FWD="$L2L2_FWD" L2L2_BACK="$L2L2_BACK" \
-    MIX_VERIFY=0 ALLOW_LATE=1 COMPOSE_PROJECT_NAME="$PROJECT" \
+    MIX_VERIFY=0 ALLOW_LATE="$ALLOW_LATE" COMPOSE_PROJECT_NAME="$PROJECT" \
     timeout 3600 "$SCRIPT_DIR/e2e-loadtest-mixed.sh" >/tmp/chaos-lt.out 2>&1
 LT_RC=$?
 say "mixed loadtest exited rc=$LT_RC"
@@ -133,7 +137,7 @@ sleep "$POST_CHAOS_SETTLE"
 say "=== (a) verify-event-completeness (legit traffic) ==="
 VC_RC=2
 if [[ -x "$TOOL_BIN" ]]; then
-    ALLOW_LATE="${ALLOW_LATE:-1}" TOOL_BIN="$TOOL_BIN" \
+    ALLOW_LATE="$ALLOW_LATE" TOOL_BIN="$TOOL_BIN" \
         NODE_CONTAINER="${PROJECT}-miden-node-1" AGGLAYER_CONTAINER="$AGGLAYER_CONTAINER" \
         "$SCRIPT_DIR/verify-event-completeness.sh" > /tmp/chaos-verify.out 2>&1
     VC_RC=$?
@@ -143,16 +147,16 @@ else
 fi
 LOCKS=$(docker logs "$AGGLAYER_CONTAINER" 2>&1 | grep -c "database is locked" || true)
 
-# ── 5a'. STORE CORROBORATION (#41) — the authoritative completeness verdict ──
-# The verifier's node-DB denominator legitimately over-counts: observed (non-
-# injected) GERs emit no UpdateHashChain, L2<->L2/reclaim claims aren't proxy-
-# sponsored, and on a RECOVERED stack the whole-history GER denominator is
-# permanently ahead of the by-design-reset log view. The proxy STORE reconciling
-# against its own authoritative sources is the real integrity signal:
-#   UHC logs == injected GERs, CLAIM logs >= landed, BRIDGE logs == emitted,
-#   and no unemitted / unbridgeable / alerted-mint rows (unclaimable is reported
-#   but non-fatal — a user-front-run leaves a benign row).
-say "=== (a') store corroboration (authoritative) ==="
+# ── 5a'. STORE CORROBORATION (#41) — ADDITIONAL failure signal, never a PASS ─
+# PR#145 review: both sides of this comparison are proxy-owned PostgreSQL
+# (synthetic_logs vs ger_entries / claim_watcher_processed /
+# bridge_out_processed) compared as aggregate counts — it cannot prove events
+# the proxy never observed, nor identity / exact-block fidelity. The
+# INDEPENDENT verifier above is the completeness authority; this section can
+# only VETO a verifier-pass (a store DROP always fails) and serves as a
+# diagnostic when the verifier's node denominator over-counts (observed GERs
+# emit no UpdateHashChain; non-sponsored claims; recovered-stack history).
+say "=== (a') store corroboration (additional failure signal) ==="
 SC_UHC=$(pgq "SELECT COUNT(*) FROM synthetic_logs WHERE topics[1] LIKE '0x65d3bf36%';")
 SC_CLAIM=$(pgq "SELECT COUNT(*) FROM synthetic_logs WHERE topics[1] LIKE '0x1df3f2a9%';")
 SC_BRIDGE=$(pgq "SELECT COUNT(*) FROM synthetic_logs WHERE topics[1] LIKE '0x50178120%';")
@@ -173,7 +177,8 @@ STORE_DROP=""
 [[ -n "${SC_BRIDGE:-}" && -n "${SC_EMIT:-}" && "${SC_BRIDGE}" -lt "${SC_EMIT}" ]] 2>/dev/null && STORE_DROP="$STORE_DROP BRIDGE<emit(${SC_BRIDGE}<${SC_EMIT})"
 if [[ -z "$STORE_DROP" ]]; then
     STORE_OK=1; say "  store corroboration: CLEAN"
-    [[ "$VC_RC" != "0" ]] && say "  (verifier mismatch with a CLEAN store = denominator artifact, not a drop)"
+    # Diagnostic label ONLY — a verifier fail still fails the verdict (PR#145).
+    [[ "$VC_RC" != "0" ]] && say "  (diagnostic: verifier mismatch with a CLEAN store is usually a denominator artifact — but the independent verifier remains authoritative)"
 else
     STORE_OK=0; say "  store corroboration: DROP —$STORE_DROP"
 fi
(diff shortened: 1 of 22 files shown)


src/store/postgres.rs
```diff
@@ -1709,16 +1711,78 @@ impl Store for PgStore {
...
+                    // live: sponsor wedged at nonce 78 until operator surgery).
+                    // Reclaim iff: `executing` AND lease expired (the owner is
+                    // contractually dead — a live admission renews its lease)
+                    // AND the bound hash was never durably admitted (no
+                    // `transactions` row ⇒ provably nothing was submitted
+                    // externally ⇒ a replacement is unambiguous). Released slots
+                    // stay permanently hash-bound (receipt identity), and an
+                    // expired `executing` slot WITH a durable row stays held —
+                    // its outcome is tracked by the pending-tx machinery.
+                    let abandoned = if state == "executing" && expired {
```
mandrigin: **[blocker] Reclaim released pre-admission failures as abandoned too**

This condition only recognizes a crashed/expired `executing` reservation as abandoned. The normal error path can leave the same provably-unadmitted slot in `released_failure`, after which a fresh-hash retry is rejected forever.

Concrete supported single-proxy interleaving:

1. Request H1 from signer A passes the first `available_capacity()` check, acquires A's per-signer lock, and wins `(A, nonce N)`.
2. A request from unrelated signer B consumes the last slot in the shared writer queue; B has a different per-signer lock.
3. H1 reaches the second capacity check in `dispatch_after_reservation` (`service_send_raw_txn.rs:714-717`) and returns `WriterQueueSaturatedError` before `txn_begin_if_absent`. No `transactions` row exists and nonce N is unchanged.
4. The outer path calls `release_reservation(..., success=false)`, changing H1's slot to `released_failure`.
5. zkevm-bridge-service claimtxman retries nonce N with freshly signed H2. Since different-hash takeover here only accepts `executing && expired`, H2 receives `HeldByOther(H1)` forever. Waiting for expiry cannot help because `released_failure` never enters this branch.

There is no ambiguous external outcome when both facts hold: the owner explicitly released failure and its hash has no durable transaction row. Durable admission happens before enqueue/Miden submission.

The KISS fix is to define a different-hash slot as abandoned when the old hash is absent from `transactions` and either `(state == "executing" && expired)` **or** `state == "released_failure"`; then bump the fence and replace it using the existing path. Mirror the predicate in `InMemoryStore`. A slot with a durable transaction row, and every `released_success` slot, must remain hash-bound.

Please add regression coverage for:

- memory store: `release_reservation(false)` with no transaction row allows a different hash immediately, with a bumped fence;
- PostgreSQL store: the identical production transition;
- both stores: `released_failure` **with** a transaction row remains `HeldByOther`;
- ideally, a service-level test that forces the second queue-capacity check to fail after reservation and proves a fresh-hash retry at the same nonce is accepted.

The existing `blocker_a_different_tx_never_takes_over_ambiguous_slot` assertion should be narrowed: a released failure without durable admission is not ambiguous.

scripts/e2e-chaos-soak.sh
```diff
@@ -169,15 +224,20 @@ say "    loadtest_rc=$LT_RC  verify_rc=$VC_RC  store_locks=$LOCKS  foreign_leak=
 # ONLY if its driver ABORTED (a fail() — e.g. a wedge or a harness bug). A
 # crashed driver means the full mixed load never ran, so it must NOT green even
 # if the (reduced) traffic verifies — else a dead driver false-passes.
-LEGIT_OK=0; [[ "$VC_RC" == "0" && "${LOCKS:-1}" == "0" && "$LT_RC" == "0" ]] && LEGIT_OK=1
+# #41: completeness = verifier PASS *or* store-corroboration CLEAN (the verifier
+# denominator over-counts by design in several benign cases); a store DROP always
+# fails regardless of the verifier.
+LEGIT_OK=0
+if [[ ( "$VC_RC" == "0" || "${STORE_OK:-0}" == "1" ) && "${LOCKS:-1}" == "0" && "$LT_RC" == "0" ]]; then LEGIT_OK=1; fi
```
mandrigin: **[blocker: release certification] Do not let proxy-owned counts override the independent completeness verifier**

This `OR` lets the highest-trust chaos soak return PASS when the only independent check has reported missing, extra, late/wrong-block, or otherwise inconsistent events. This is a test-certification blocker rather than evidence of a separate runtime defect.

`STORE_OK` is not an alternative source of truth: both sides come from proxy-owned PostgreSQL tables (`synthetic_logs` versus `ger_entries`, `claim_watcher_processed`, and `bridge_out_processed`), and they are compared only as aggregate counts. If the proxy never observes a consumed node note, both its processed row and its synthetic log are absent, so the store still reports CLEAN. Conversely, an extra synthetic log is accepted because the checks only fail on `logs < rows`; identity and exact block are not checked. Therefore `VC_RC=1, STORE_OK=1` currently becomes `LEGIT_OK=1`.

This also invalidates private/tag-0 containment: lines 215-216 explicitly rely on the verifier's `extra==0`, but `GARBO_OK` never requires `VC_RC==0`. A leaked private-derived event can fail the verifier while these aggregate store checks remain CLEAN, and the overall soak can still pass. The direct foreign-GI check is stronger and should remain.

There is a second false-green feeding this verdict: this script launches `e2e-loadtest-mixed.sh` with `MIX_VERIFY=0`, and that script exits 0 at lines 416-419 **before** enforcing its L1 return code, forward/back landed counts, or address-clash result at lines 442-448. Thus `LT_RC=0` means only that the driver reached the shortcut, not that every requested operation landed. An operation that was never submitted or consumed is also absent from the later node denominator, so the post-heal verifier cannot reconstruct that missing workload assertion.

Minimal fix:

1. Require `VC_RC == 0` for `LEGIT_OK`; retain store corroboration as an additional failure signal/diagnostic, never as an override.
2. Require the independent verifier, or a direct persisted-private-note-ID absence assertion, for private garbo containment.
3. Make `MIX_VERIFY=0` skip only the duplicate verifier invocation while still executing the mixed test's L1/fwd/back/clash/lock verdict. Alternatively export structured operation counts and assert them here.
4. If PASS is meant to certify **exact-block** fidelity, run the verifier with `ALLOW_LATE=0`; otherwise weaken the PASS wording because the current value of 1 permits late events.

Please add a small verdict regression test (a factored shell predicate is sufficient) proving at least:

- `VC_RC=1, STORE_OK=1` fails;
- a private-derived extra event fails;
- incomplete forward/back counts or a non-distinct clash returns nonzero even with `MIX_VERIFY=0`;
- only complete operations plus `VC_RC=0` can pass.

mandrigin (commented):
Architecture-mapped re-review: these are the two remaining actionable issues for the supported single-proxy deployment and its release-certification harness. I am intentionally not requesting distributed/multi-replica fencing here; that remains out of scope.

src/store/postgres.rs
mandrigin: Addressed in `64a5824`.

Re-reviewed against the supported single-proxy flow:

- both PostgreSQL and the in-memory store now classify a different-hash slot as abandoned when there is no durable transaction row and the state is either expired `executing` or `released_failure`;
- reclamation immediately replaces the hash and bumps the fence;
- `released_success` and every slot with a durable transaction row remain hash-bound;
- coverage includes unadmitted/admitted `released_failure` in both stores plus a service-level fresh-hash retry at the same nonce.

This is the minimal intended fix; no distributed/multi-replica admission redesign was added. Resolving this thread.

scripts/e2e-chaos-soak.sh
mandrigin: Re-reviewed at `c6f42bb`. The original store/verifier override, private containment, outer `MIX_VERIFY=0` shortcut, and strict-by-default `ALLOW_LATE=0` are addressed. One source-level false-green remains in the nested L1↔Miden workload, so this thread should stay open.

`e2e-loadtest-mixed.sh:259-261` launches `e2e-bridge-loadtest-isolated.sh` with `VERIFY=0` and treats that child's status as `LT_RC`. The child records submission failures in `TOT_FAIL_L1/TOT_FAIL_L2` and computes `G_SUB/G_CLM`, but its final status at `e2e-bridge-loadtest-isolated.sh:680` is only:

```bash
exit $(( LOCK_COUNT > 0 ? 1 : VERIFY_RC ))
```

With `VERIFY=0`, `VERIFY_RC` remains zero. Therefore L1→Miden or Miden→L1 submissions may fail—or successful submissions may remain unclaimed—while the child returns 0 as long as there are no database-lock errors. The parent then sees `LT_RC=0`; its new `mixed_ops_ok` only validates the separate L2B legs/clash, and the post-heal completeness verifier cannot detect an intended operation that was never submitted/consumed. The advertised four-direction soak can still PASS with missing L1 workload.

Minimal KISS fix:

1. Add a strict operational-result mode to `e2e-bridge-loadtest-isolated.sh`, enabled by the mixed runner.
2. In that mode return nonzero unless all of these hold:
   - `TOT_SUB_L1 == PLAN_L1_TARGET`;
   - `TOT_SUB_L2 == PLAN_L2_TARGET`;
   - `TOT_FAIL_L1 == 0 && TOT_FAIL_L2 == 0`;
   - `G_CLM == G_SUB`.
3. Combine that operational result with the existing lock/verifier result at the child's final exit. Keeping the check in the child avoids parsing human-readable logs in the parent.
4. Extend the factored predicate tests so target shortfall, explicit submission failure, and submitted-but-unclaimed work all produce the real child operational failure consumed as `LT_RC`.

The current verdict-unit test injects an artificial `LT_RC=1`; it proves the parent rejects a failure, but not that this child turns incomplete L1 work into that failure.

Non-blocking wording nit: when an operator explicitly sets `ALLOW_LATE=1`, the final PASS line should not still claim `exact-block` fidelity; make that label conditional on the mode.

mandrigin: Addressed in `5968603`.

Source-level re-review confirms the full status chain now fails closed:

- the mixed runner enables `STRICT_OPS=1` and supplies exact directional targets;
- the nested L1↔Miden runner compares successful submissions with each planned target, requires zero explicit submission failures, and requires `G_CLM == G_SUB`;
- any target shortfall, failed submission, or submitted-but-unclaimed operation sets the child's real exit status nonzero even with `VERIFY=0`;
- the parent captures that status as `LT_RC`, `mixed_ops_ok` requires zero, and the final chaos verdict also requires the mixed runner to pass;
- `ALLOW_LATE=1` now produces truthful non-exact-block PASS wording.

The changed scripts pass `bash -n`, and the factored verdict suite passes the target-shortfall, explicit-failure, unclaimed-work, empty-input, and complete-success cases. No remaining false-green was found in the reviewed mixed/chaos path. Full Docker soak was not part of this source re-review.

`STRICT_OPS` remains opt-in for standalone reliability/measurement runs; the mixed/chaos certification path enables it explicitly, which is the intended scope here. Resolving this thread.

IvanBelyakoff (approved):

PR #150, opened 2026-07-21
fix(admin): validate native-faucet metadata against deployed account

### LINKED ISSUE #149: fix(admin): validate native-faucet metadata at registration time
## Context

Follow-up from [PR #145](https://github.com/gateway-fm/miden-agglayer/pull/145). This is intentionally separate from the release patch.

Recovery now reads an independent Miden faucet account's authoritative `token_name`, `symbol`, and `decimals`. However, `admin_registerNativeFaucet` currently trusts caller-supplied metadata and may default an omitted name to the supplied symbol.

If the registered values differ from the already-deployed faucet account, the proxy can persist and emit a metadata hash whose preimage cannot later be reconstructed from authoritative chain state after database loss.

Custom `name != symbol` faucets are valid and must be preserved exactly.

## Minimal scope

- During native-faucet registration, import/read the referenced public faucet account from Miden.
- Compare its actual token name, symbol, and decimals with the requested values before inserting the registry row or emitting bridge metadata.
- Reject any mismatch with a precise operator-facing error and no partial state change.
- Preserve an authentic custom name when it differs from the symbol; do not normalize it away.
- Define how operators detect and remediate already-registered mismatched rows. Recovery must not silently guess a metadata preimage.

Do not redesign general metadata recovery or add distributed-replica coordination.

## Acceptance criteria

- Registration of a faucet whose requested name, symbol, and decimals exactly match the Miden account succeeds.
- A matching custom-name faucet (`name != symbol`) survives database-loss restore with the same metadata hash/preimage.
- A mismatched name, symbol, or decimals is rejected independently with a specific error before the registry changes.
- If `name` is omitted, registration succeeds only when the faucet's actual name equals the supplied symbol, unless the API imports and persists the actual name automatically.
- A regression proves failed registration leaves no registry row and emits no bridge configuration metadata.
- An E2E restore explicitly verifies both the native-faucet row and a network-2 token registry row after recovery.
- Operator documentation explains detection and remediation of any legacy mismatched row without silently rewriting history.

## Non-goals

- Accepting caller metadata as authoritative over the deployed faucet account.
- Normalizing every valid token name to its symbol.
- Active-active/shared-PostgreSQL proxy replicas; tracked separately in [#142](https://github.com/gateway-fm/miden-agglayer/issues/142).

mandrigin:
## Summary

- read the referenced deployed Miden faucet before native-faucet registration
- validate the requested name, symbol, and decimals against the faucet's authoritative metadata
- persist the authoritative metadata preimage, including valid custom names that differ from the symbol
- add unit and Miden-origin E2E coverage for mismatch rejection, custom-name preservation, and database-loss restore
- document how operators can detect and remediate legacy mismatched registry rows

## Why

`admin_registerNativeFaucet` previously trusted caller-supplied metadata. If those values differed from the deployed faucet account, the proxy could persist a metadata-hash preimage that could not be reconstructed from chain state during database-loss recovery.

This change makes the deployed faucet authoritative and rejects mismatches before bridge configuration or registry persistence.

## Test coverage added

- unit tests for exact matches, independent field mismatches, omitted names, and custom `name != symbol`
- fail-closed registration test when authoritative faucet metadata cannot be read
- Miden-origin E2E assertions for mismatch rejection and custom-name persistence
- opt-in destructive restore regression covering both a native custom-name row and a network-2 row

Closes #149

### DIFF
--- a/docs/operations/runbook.md
+++ b/docs/operations/runbook.md
@@ -319,17 +319,17 @@ leaf). To detect it:
    faucet's on-chain `MetadataHash`. A row whose stored preimage keccak differs
    from the deployed faucet's hash is mismatched.
 
-2. Remediate. Re-registration does **not** repair a mismatched row:
-   `admin_registerNativeFaucet` is register-if-absent (idempotent) — for an
-   origin that already has a route it returns the existing row unchanged and
-   never rewrites its metadata. A kept stack with a mismatched legacy row must
-   therefore be remediated by removing that row (operator-authorized DB action)
-   and re-registering the faucet against its deployed account — the fresh
-   registration then validates + persists authoritative metadata (supplying the
-   faucet's actual name, or omitting `name` to adopt it; a custom `name != symbol`
-   is preserved, never normalized). Recovery must not reconstruct history by
-   guessing; the only safe source is the deployed faucet account. On the
-   supported clean-slate rollout this situation does not arise.
+2. Do not attempt an in-place repair. `admin_registerNativeFaucet` is
+   register-if-absent (idempotent) — it never rewrites an existing row — so
+   re-registration cannot fix a mismatched row. Surgical row deletion +
+   re-registration is also **not** a repair: it discards the only locally
+   retained legacy preimage and overwrites the bridge's current metadata hash,
+   yet it cannot repair historical B2AGG leaves/events already committed with
+   the old hash. The supported rollout is clean-slate, so legacy state is not
+   migrated and this situation does not arise on it. If a mismatched row is
+   ever observed on a stack that must be kept: preserve and back up the current
+   state, quarantine the affected faucet, and escalate — or rebuild from a
+   clean deployment. Do not perform surgical row edits.
 
 ## 4. Incident procedures
 
--- a/docs/operations/runbook.md
+++ b/docs/operations/runbook.md
@@ -291,10 +291,16 @@ preimage is reconstructable from authoritative chain state during `--restore`
 mismatched symbol, decimals, or name is rejected up-front with a specific error
 and leaves no registry row.
 
+**Legacy state is not migrated.** The supported rollout is clean-slate: the new
+deployment starts fresh with this validation active, so no legacy mismatched row
+carries into it, and no in-place repair path is provided. The detection below is
+for diagnosing an unexpected mismatched row on a stack that must be kept — not a
+supported migration.
+
 A row registered by an **older build** may still carry a preimage that does not
 match its deployed faucet account. Recovery does **not** silently guess a
 preimage — an unrecoverable native row halts `--restore` fail-closed (its poison
-leaf). To detect and remediate before that happens:
+leaf). To detect it:
 
 1. Detect. For each native row (`origin_network` == the proxy's configured
    `network_id`), compare the stored preimage against the deployed faucet
@@ -313,12 +319,17 @@ leaf). To detect and remediate before that happens:
    faucet's on-chain `MetadataHash`. A row whose stored preimage keccak differs
    from the deployed faucet's hash is mismatched.
 
-2. Remediate. Re-register the faucet against its deployed account so the row and
-   the emitted bridge metadata are rewritten from authoritative state. Because a
-   custom `name != symbol` is valid, supply the faucet's actual name (or omit
-   `name` to adopt it) — never normalize it to the symbol. Recovery must not
-   reconstruct history by guessing; the only safe source is the deployed faucet
-   account.
+2. Remediate. Re-registration does **not** repair a mismatched row:
+   `admin_registerNativeFaucet` is register-if-absent (idempotent) — for an
+   origin that already has a route it returns the existing row unchanged and
+   never rewrites its metadata. A kept stack with a mismatched legacy row must
+   therefore be remediated by removing that row (operator-authorized DB action)
+   and re-registering the faucet against its deployed account — the fresh
+   registration then validates + persists authoritative metadata (supplying the
+   faucet's actual name, or omitting `name` to adopt it; a custom `name != symbol`
+   is preserved, never normalized). Recovery must not reconstruct history by
+   guessing; the only safe source is the deployed faucet account. On the
+   supported clean-slate rollout this situation does not arise.
 
 ## 4. Incident procedures
 
--- a/docs/operations/runbook.md
+++ b/docs/operations/runbook.md
@@ -280,6 +280,46 @@ sweep.
 Never substitute `--init` for recovery: it creates new account identities and
 can strand control/balances associated with the old ones.
 
+### Detecting + remediating a mismatched native-faucet registry row
+
+`admin_registerNativeFaucet` now validates caller-supplied metadata against the
+deployed Miden faucet account before writing anything (issue #149): the persisted
++ emitted metadata-hash preimage `abi.encode(name, symbol, decimals)` is taken
+from the faucet account, never from caller-supplied params. This guarantees the
+preimage is reconstructable from authoritative chain state during `--restore`
+(recovery derives its only native-token candidate from the faucet account). A
+mismatched symbol, decimals, or name is rejected up-front with a specific error
+and leaves no registry row.
+
+A row registered by an **older build** may still carry a preimage that does not
+match its deployed faucet account. Recovery does **not** silently guess a
+preimage — an unrecoverable native row halts `--restore` fail-closed (its poison
+leaf). To detect and remediate before that happens:
+
+1. Detect. For each native row (`origin_network` == the proxy's configured
+   `network_id`), compare the stored preimage against the deployed faucet
+   account's authoritative `token_name` / `symbol` / `decimals`:
+
+   ```sql
+   -- stored preimage (hex) per native faucet
+   SELECT faucet_id, symbol, origin_decimals, encode(metadata,'hex')
+   FROM faucet_registry
+   WHERE origin_network = <configured network_id>;
+   ```
+
+   Read the faucet account's authoritative metadata from Miden (the same
+   `token_name()` / `symbol()` / `decimals()` the proxy reads at registration),
+   ABI-encode `(name, symbol, decimals)`, and confirm its `keccak256` equals the
+   faucet's on-chain `MetadataHash`. A row whose stored preimage keccak differs
+   from the deployed faucet's hash is mismatched.
+
+2. Remediate. Re-register the faucet against its deployed account so the row and
+   the emitted bridge metadata are rewritten from authoritative state. Because a
+   custom `name != symbol` is valid, supply the faucet's actual name (or omit
+   `name` to adopt it) — never normalize it to the symbol. Recovery must not
+   reconstruct history by guessing; the only safe source is the deployed faucet
+   account.
+
 ## 4. Incident procedures
 
 ### Node outage or `/health` 503
--- a/scripts/e2e-cantina13-metadata-recovery.sh
+++ b/scripts/e2e-cantina13-metadata-recovery.sh
@@ -439,16 +439,26 @@ pass "DB backed up ($(wc -c < "$BACKUP") bytes → $BACKUP)"
 # reuses rows that already exist in the full suite — no extra registration, no separate
 # proxy-only reset (PR #150 re-review). If either row is absent (cantina13 run outside
 # the full suite), SKIP — the assertion is only meaningful with the prior tiers present.
-_s149_row() { pg "SELECT faucet_id||'|'||origin_network||'|'||encode(origin_address,'hex')||'|'||symbol||'|'||origin_decimals||'|'||encode(metadata,'hex') FROM faucet_registry WHERE $1 ORDER BY faucet_id LIMIT 1"; }
-S149_NATIVE_ROW=$(_s149_row "symbol='WMDN'")
+# FULL identity: faucet_id | origin_network | origin_address | symbol |
+# origin_decimals | miden_decimals | scale | metadata-preimage — miden_decimals
+# and scale included so a restore with wrong CONVERSION params can't pass (PR #150).
+_s149_row() { pg "SELECT faucet_id||'|'||origin_network||'|'||encode(origin_address,'hex')||'|'||symbol||'|'||origin_decimals||'|'||miden_decimals||'|'||scale||'|'||encode(metadata,'hex') FROM faucet_registry WHERE $1 ORDER BY faucet_id LIMIT 1"; }
+# Constrain the native custom-name capture to the NATIVE route (origin_network ==
+# the proxy's own net id): symbol='WMDN' alone could select a foreign-origin WMDN
+# row on a warm/shared stack and prove survival of the WRONG route (PR #150).
+S149_NATIVE_ROW=$(_s149_row "symbol='WMDN' AND origin_network=${LOCAL_NETWORK_ID}")
 S149_NET2_ROW=$(_s149_row "origin_network=2")
 if [[ -n "$S149_NATIVE_ROW" && -n "$S149_NET2_ROW" ]]; then
     S149_CHECK=1
     S149_NATIVE_FID="${S149_NATIVE_ROW%%|*}"; S149_NET2_FID="${S149_NET2_ROW%%|*}"
     pass "#149 restore-survival PRE: captured native custom-name row (WMDN $S149_NATIVE_FID) + net-2 row ($S149_NET2_FID)"
+elif [[ "${REQUIRE_S149_RESTORE:-0}" == "1" ]]; then
+    # In the full overlay suite the prior L2L2/native phases MUST have created both
+    # rows; a regression there must FAIL the release gate, not silently skip (PR #150).
+    fail "#149 restore-survival: REQUIRE_S149_RESTORE=1 but a prerequisite row is absent (native WMDN net=$LOCAL_NETWORK_ID: ${S149_NATIVE_ROW:-<none>} | net-2: ${S149_NET2_ROW:-<none>}) — a prior phase regressed"
 else
     S149_CHECK=0
-    log "#149 restore-survival: SKIP — no WMDN native and/or net-2 row present (cantina13 run outside the full suite)"
+    log "#149 restore-survival: SKIP — no WMDN native and/or net-2 row present (standalone cantina13, REQUIRE_S149_RESTORE unset)"
 fi
 
 # finding #65 — we deliberately do NOT preserve the proxy's per-signer nonce tables.
--- a/scripts/e2e-cantina13-metadata-recovery.sh
+++ b/scripts/e2e-cantina13-metadata-recovery.sh
@@ -446,8 +446,14 @@ _s149_row() { pg "SELECT faucet_id||'|'||origin_network||'|'||encode(origin_addr
 # Constrain the native custom-name capture to the NATIVE route (origin_network ==
 # the proxy's own net id): symbol='WMDN' alone could select a foreign-origin WMDN
 # row on a warm/shared stack and prove survival of the WRONG route (PR #150).
-S149_NATIVE_ROW=$(_s149_row "symbol='WMDN' AND origin_network=${LOCAL_NETWORK_ID}")
-S149_NET2_ROW=$(_s149_row "origin_network=2")
+# REQUIRE non-empty metadata (octet_length > 0): a row whose preimage is empty
+# would still yield a non-empty _s149_row tuple, and the POST equality would accept
+# empty->empty — a false-green that never exercises recovery of the preimage (for
+# net-2, recovery from its OWN network RPC, finding #62). An empty-metadata row is
+# treated as ABSENT here, so the absent-row policy below (fail if REQUIRE, else skip)
+# applies (PR #150 re-review).
+S149_NATIVE_ROW=$(_s149_row "symbol='WMDN' AND origin_network=${LOCAL_NETWORK_ID} AND octet_length(metadata) > 0")
+S149_NET2_ROW=$(_s149_row "origin_network=2 AND octet_length(metadata) > 0")
 if [[ -n "$S149_NATIVE_ROW" && -n "$S149_NET2_ROW" ]]; then
     S149_CHECK=1
     S149_NATIVE_FID="${S149_NATIVE_ROW%%|*}"; S149_NET2_FID="${S149_NET2_ROW%%|*}"
@@ -455,7 +461,7 @@ if [[ -n "$S149_NATIVE_ROW" && -n "$S149_NET2_ROW" ]]; then
 elif [[ "${REQUIRE_S149_RESTORE:-0}" == "1" ]]; then
     # In the full overlay suite the prior L2L2/native phases MUST have created both
     # rows; a regression there must FAIL the release gate, not silently skip (PR #150).
-    fail "#149 restore-survival: REQUIRE_S149_RESTORE=1 but a prerequisite row is absent (native WMDN net=$LOCAL_NETWORK_ID: ${S149_NATIVE_ROW:-<none>} | net-2: ${S149_NET2_ROW:-<none>}) — a prior phase regressed"
+    fail "#149 restore-survival: REQUIRE_S149_RESTORE=1 but a prerequisite row is absent or has empty metadata (native WMDN net=$LOCAL_NETWORK_ID: ${S149_NATIVE_ROW:-<none>} | net-2: ${S149_NET2_ROW:-<none>}) — a prior phase regressed"
 else
     S149_CHECK=0
     log "#149 restore-survival: SKIP — no WMDN native and/or net-2 row present (standalone cantina13, REQUIRE_S149_RESTORE unset)"
--- a/scripts/e2e-cantina13-metadata-recovery.sh
+++ b/scripts/e2e-cantina13-metadata-recovery.sh
@@ -430,6 +430,27 @@ docker exec "$PG_CONTAINER" pg_dump -U agglayer agglayer_store > "$BACKUP" 2>/de
 [[ -s "$BACKUP" ]] || fail "DB backup failed (empty dump) — refusing to drop without a backup"
 pass "DB backed up ($(wc -c < "$BACKUP") bytes → $BACKUP)"
 
+# ── #149 restore-survival PRE-capture (reuses existing full-suite rows) ────────────────
+# Before the from-scratch DROP + --restore, snapshot a native CUSTOM-NAME row and a
+# network-2 row so we can prove they SURVIVE recovery with FULL identity
+# (faucet_id | origin_network | origin_address | symbol | origin_decimals | preimage).
+# The native custom-name (WMDN, name != symbol) row comes from e2e-miden-origin.sh's
+# custom-name faucet (step 1d); a net-2 (L2B-origin) row comes from the l2l2 tier. This
+# reuses rows that already exist in the full suite — no extra registration, no separate
+# proxy-only reset (PR #150 re-review). If either row is absent (cantina13 run outside
+# the full suite), SKIP — the assertion is only meaningful with the prior tiers present.
+_s149_row() { pg "SELECT faucet_id||'|'||origin_network||'|'||encode(origin_address,'hex')||'|'||symbol||'|'||origin_decimals||'|'||encode(metadata,'hex') FROM faucet_registry WHERE $1 ORDER BY faucet_id LIMIT 1"; }
+S149_NATIVE_ROW=$(_s149_row "symbol='WMDN'")
+S149_NET2_ROW=$(_s149_row "origin_network=2")
+if [[ -n "$S149_NATIVE_ROW" && -n "$S149_NET2_ROW" ]]; then
+    S149_CHECK=1
+    S149_NATIVE_FID="${S149_NATIVE_ROW%%|*}"; S149_NET2_FID="${S149_NET2_ROW%%|*}"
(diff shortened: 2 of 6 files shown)


src/service_admin.rs
```diff
@@ -325,23 +407,88 @@ pub async fn admin_register_native_faucet(
         return Ok(existing.faucet_id.to_hex());
```
mandrigin, line 407: **Updated — non-blocking under the confirmed clean-slate rollout.** The 0.15.9 deployment and its state will be removed, so no legacy poisoned row can carry into the new testnet and no repair implementation is required here. The endpoint's early return is therefore harmless for already-valid rows. Please only remove/reword the runbook claim that ordinary re-registration repairs legacy state; the supported policy is that legacy state is not migrated.

src/service_admin.rs
```diff
@@ -325,23 +407,88 @@ pub async fn admin_register_native_faucet(
...
+                    .await
+                    .map_err(|e| anyhow::anyhow!("get_account({faucet_id}): {e}"))?
+                    .ok_or_else(|| {
+                        anyhow::anyhow!(
+                            "admin_registerNativeFaucet: faucet account {faucet_id} not \
+                             found after import"
+                        )
+                    })?;
+                // Both supported kinds expose the standard FungibleFaucet metadata.
+                let (_kind, faucet) = faucet_ops::classify_faucet_account(&faucet_account)?;
```
mandrigin: **Updated — non-blocking, but recommended.** Miden restricts this configuration to the bridge admin and the native path can only lock/unlock existing assets, so this is not an unbacked-mint or external-user vulnerability. Still, retaining `kind` and requiring `FaucetKind::NativeFungible` is a very small, useful guard against an operator accidentally reclassifying an already-registered AggLayer-owned faucet. Full hostile-account hardening need not expand #149.

scripts/e2e-miden-origin.sh
```diff
@@ -126,6 +126,84 @@ done
...
+      \"symbol\":\"$1\",\"decimals\":$2}]}" 2>/dev/null
+}
+BAD_SYM_ORIGIN="0x0bad51$(python3 -c "import secrets;print(secrets.token_hex(17))")"
+BAD_DEC_ORIGIN="0x0badde$(python3 -c "import secrets;print(secrets.token_hex(17))")"
+BAD_NAME_ORIGIN="0x0bad1a$(python3 -c "import secrets;print(secrets.token_hex(17))")"
+# wrong symbol
+RESP=$(_reg_mismatch "WRONG" 8 "$BAD_SYM_ORIGIN")
+echo "$RESP" | grep -qi 'symbol mismatch' \
+  || fail "#149: wrong-symbol registration should error 'symbol mismatch', got: $RESP"
+[[ -z "$(pgq "SELECT 1 FROM faucet_registry WHERE lower(origin_address)=lower('${BAD_SYM_ORIGIN#0x}') OR lower(origin_address)=lower('$BAD_SYM_ORIGIN');")" ]] \
```
mandrigin: **Blocking test false-green:** `origin_address` is `BYTEA`, so PostgreSQL has no `lower(origin_address)`. `pgq` suppresses the SQL error and this `[[ -z "$(...)" ]]` observes empty output, allowing the assertion to pass even if a row exists. Compare `encode(origin_address, 'hex')` with `${BAD_SYM_ORIGIN#0x}` (or use `decode`). Please also make the no-row/no-config assertion for decimals and name; line 163 currently claims all three despite checking only symbol, and the acceptance criterion requires no bridge metadata emission as well as no row.

scripts/e2e-miden-origin.sh
```diff
@@ -231,6 +309,92 @@ WRAPPED_SUPPLY=$(cast call "$WRAPPED_L2B" "totalSupply()(uint256)" --rpc-url "$D
...
 pass "NET-ZERO: native holder = $NATIVE_BAL units; wrapped $DEST_LABEL supply = 0"
 
+# ── 1e. #149 — --restore preserves the native custom-name row AND a network-2 row ─
+# The restore side of #149: a custom-name (name != symbol) native faucet's metadata
+# preimage must be reconstructable from authoritative chain state after DB loss (it is,
+# because #149 persists the authoritative name), and multi-network recovery must rebuild
+# a network-2 (L2B-origin) row too. Register a net-2 token, drop + rebuild the proxy
+# store from on-chain (same runbook as e2e-cantina13-metadata-recovery.sh), and assert
+# BOTH rows survive with byte-identical metadata preimages. l2b-only (net-2 needs L2B).
+if [[ "$DEST" == "l2b" && "${RUN_1E_RESTORE:-0}" == "1" ]]; then
```
mandrigin: **Blocking coverage gap:** no repository caller sets `RUN_1E_RESTORE=1`; the normal `e2e-test.sh` calls both Miden-origin runs without it. Therefore the only custom-name + network-2 restore proof is dead code and #149's explicit restore acceptance criterion is never exercised. Please wire a dedicated one-shot invocation into the real gate at a safe final position.

scripts/e2e-miden-origin.sh
```diff
@@ -231,6 +309,92 @@ WRAPPED_SUPPLY=$(cast call "$WRAPPED_L2B" "totalSupply()(uint256)" --rpc-url "$D
...
+    # Add a network-2 (L2B-origin) wrapped-token row so a net-2 identity exists to recover.
+    NET2_ORIGIN="0x0e2b02$(python3 -c "import secrets;print(secrets.token_hex(17))")"
+    RESP=$(curl -sf "$L2_RPC" -H "Content-Type: application/json" -H "$ADMIN_BEARER" -d "{
+      \"jsonrpc\":\"2.0\",\"id\":1,\"method\":\"admin_registerFaucet\",
+      \"params\":[{\"symbol\":\"L2BT\",\"origin_token_address\":\"$NET2_ORIGIN\",
+        \"origin_network\":2,\"origin_decimals\":6}]}" 2>/dev/null) \
+      || fail "#149/1e: admin_registerFaucet (net-2) unreachable: $RESP"
+    NET2_FAUCET_ID=""
+    for _i in $(seq 1 40); do
+        NET2_FAUCET_ID=$(pgq "SELECT faucet_id FROM faucet_registry WHERE origin_network=2 AND (lower(origin_address)=lower('${NET2_ORIGIN#0x}') OR lower(origin_address)=lower('$NET2_ORIGIN'));")
```
mandrigin: **Blocking when the gate is enabled:** this is the same invalid `lower(BYTEA)` comparison. Here it is a plain assignment under `set -e`, so the opt-in restore scenario exits on its first poll and never reaches restore. Use `encode(origin_address, 'hex') = lower('${NET2_ORIGIN#0x}')` or a `decode(..., 'hex')` bytea comparison.

mandrigin (commented):
Updated after confirming the deployment model:

- **Legacy migration is not a blocker.** The 0.15.9 deployment and its state will be removed; the new testnet starts clean with this validation active. No repair path for old rows is required. The inaccurate runbook remediation should simply be removed or state that legacy state is not migrated.
- **Faucet-kind validation is recommended but non-blocking.** Miden restricts configuration to the bridge admin and preserves supply through lock/unlock semantics. Still, checking `FaucetKind::NativeFungible` is a small, useful operator-error guard and should remain in this patch if possible.

The remaining merge blockers are the E2E defects: invalid `lower(BYTEA)` comparisons (including the false-green no-row assertion), incomplete mismatch side-effect assertions, and the restore regression not being invoked by any normal gate.

scripts/e2e-miden-origin.sh
mandrigin: Partially addressed in `40bea59`: the BYTEA expression is now valid and symbol/decimals/name each get a no-row check. The remaining DB-error false-green and mismatch-path no-config proof are captured in new comments on the current lines, so this outdated combined thread can be resolved.

scripts/e2e-miden-origin.sh
mandrigin: Addressed in `40bea59`: the net-2 poll now compares `encode(origin_address,'hex')`, so the restore leg no longer exits on `lower(BYTEA)`. Resolving.

scripts/e2e-miden-origin.sh
mandrigin: The dead-code part is addressed in `40bea59`: `e2e-test.sh all` now invokes `RUN_1E_RESTORE=1`. Resolving this old thread; the new invocation's reset ordering and stale external nonce state are a separate current-line blocker.

src/service_admin.rs
mandrigin: Addressed in `40bea59`: the endpoint now retains `FaucetKind` and rejects `AggLayerOwned` before bridge or store mutation. This is the small defensive check requested; resolving.

docs/operations/runbook.md
```diff
@@ -280,6 +280,57 @@ sweep.
...
+   `token_name()` / `symbol()` / `decimals()` the proxy reads at registration),
+   ABI-encode `(name, symbol, decimals)`, and confirm its `keccak256` equals the
+   faucet's on-chain `MetadataHash`. A row whose stored preimage keccak differs
+   from the deployed faucet's hash is mismatched.
+
+2. Remediate. Re-registration does **not** repair a mismatched row:
+   `admin_registerNativeFaucet` is register-if-absent (idempotent) — for an
+   origin that already has a route it returns the existing row unchanged and
+   never rewrites its metadata. A kept stack with a mismatched legacy row must
+   therefore be remediated by removing that row (operator-authorized DB action)
```
mandrigin: **Blocking documentation safety:** this section first says no in-place legacy repair is supported, then prescribes deleting one live row and re-registering. That discards the only locally retained legacy preimage and overwrites the bridge's current metadata hash, but it cannot repair historical B2AGG leaves/events committed with the old hash. Given the confirmed clean-slate rollout, the minimal safe documentation is: legacy state is not migrated; preserve/backup, quarantine, and escalate or rebuild—do not recommend surgical row deletion/re-registration.

scripts/e2e-miden-origin.sh
```diff
@@ -126,6 +126,116 @@ done
...
+    \"params\":[{\"faucet_id\":\"$NATIVE_FAUCET_ID\",\"origin_token_address\":\"$3\",
+      \"symbol\":\"$1\",\"decimals\":$2}]}" 2>/dev/null
+}
+BAD_SYM_ORIGIN="0x0bad51$(python3 -c "import secrets;print(secrets.token_hex(17))")"
+BAD_DEC_ORIGIN="0x0badde$(python3 -c "import secrets;print(secrets.token_hex(17))")"
+BAD_NAME_ORIGIN="0x0bad1a$(python3 -c "import secrets;print(secrets.token_hex(17))")"
+# origin_address is BYTEA: `lower(origin_address)` is an invalid Postgres call that
+# pgq suppresses → empty output → a FALSE-GREEN "no row" even if one exists (PR #150
+# review). Compare `encode(origin_address,'hex')` against the hex origin instead.
+_no_registry_row() { [[ -z "$(pgq "SELECT 1 FROM faucet_registry WHERE encode(origin_address,'hex')=lower('${1#0x}');")" ]]; }
```
mandrigin: **Blocking false-green remains:** the BYTEA expression is fixed, but any `pgq`/connection/query failure still produces empty output (stderr is suppressed) and therefore passes this `-z` check. Capture `SELECT COUNT(*)`, propagate a nonzero `pgq` exit, and require the result to equal exactly `0`; empty/error must fail.

scripts/e2e-miden-origin.sh
```diff
@@ -126,6 +126,116 @@ done
...
+RESP=$(curl -s "$L2_RPC" -H "Content-Type: application/json" -H "$ADMIN_BEARER" -d "{
+  \"jsonrpc\":\"2.0\",\"id\":1,\"method\":\"admin_registerNativeFaucet\",
+  \"params\":[{\"faucet_id\":\"$NATIVE_FAUCET_ID\",\"origin_token_address\":\"$BAD_NAME_ORIGIN\",
+    \"symbol\":\"MDN\",\"decimals\":8,\"name\":\"Not The Real Name\"}]}" 2>/dev/null)
+echo "$RESP" | grep -qi 'name mismatch' \
+  || fail "#149: wrong-name registration should error 'name mismatch', got: $RESP"
+_no_registry_row "$BAD_NAME_ORIGIN" || fail "#149: wrong-name registration must leave NO registry row"
+# No-bridge-metadata side: register_faucet_in_bridge runs only AFTER validation passes,
+# so a rejected registration also emits NO ConfigAggBridgeNote — unit-proven by
+# native_registration_fails_closed_when_metadata_unreadable (test_call_count()==1).
```
mandrigin: **Blocking acceptance evidence:** the cited unit test covers an unreadable account, not a successful authoritative read followed by a name/symbol/decimals mismatch; its test stub does not execute the closure or materialize an account. The E2E checks only PostgreSQL. Add a mismatch-path test that proves the second client/config call is absent, or assert the fresh origin is absent from the bridge's on-chain config/token map. Please do not label this `unit-proven` until that path is exercised.

scripts/e2e-test.sh
```diff
@@ -88,6 +88,16 @@ case "$test_filter" in
...
+            echo ""
+            # #149 restore acceptance: a dedicated one-shot proving a native CUSTOM-NAME
+            # faucet (name != symbol) AND a network-2 row survive a from-scratch drop +
+            # --restore with byte-identical metadata preimages (RUN_1E_RESTORE=1). It is
+            # DESTRUCTIVE (rebuilds the store from chain) so it runs as its own phase,
+            # immediately before cantina13 (which does its own from-scratch restore next,
+            # so back-to-back restores are safe). Without this wiring the restore leg is
+            # dead code — PR #150 review.
+            echo "== #149 native-faucet metadata restore-survival (drop + --restore) =="
+            RUN_1E_RESTORE=1 DEST=l2b "$SCRIPT_DIR/e2e-miden-origin.sh"
```
mandrigin: **Blocking runner ordering:** this invocation drops the proxy DB/reset nonce state but restarts only `miden-agglayer`; Cantina13 then drives L1→Miden autoclaims before its own coordinated bridge-service reset. Cantina13 documents that the claimsponsor's cached nonce is stale after a proxy-only reset and will wedge. Also, this is the third native run on the same proxy, so moving net-2 registration to the start of this script does not provide the `fresh client` claimed for finding #76. Use a dedicated disposable phase/job, or explicitly coordinate the external bridge-service reset and a fresh proxy; reusing an already-existing net-2 row would avoid the extra registration entirely.

scripts/e2e-miden-origin.sh
```diff
@@ -231,6 +341,79 @@ WRAPPED_SUPPLY=$(cast call "$WRAPPED_L2B" "totalSupply()(uint256)" --rpc-url "$D
...
+    wait_for "proxy healthy after 1e rebuild" \
+        "[[ \$(docker inspect -f '{{.State.Health.Status}}' $AGGLAYER_CONTAINER 2>/dev/null) == healthy ]]" 180 3
+    pass "1e: store rebuilt from on-chain"
+
+    # POST: BOTH rows survive with byte-identical metadata preimages (name != symbol
+    # preserved on the native row). Restore rebuilds faucet identities + metadata eagerly
+    # (rebuild_faucet_entry_from_chain), so poll for the rows then compare.
+    POST_NATIVE_META=""; POST_NET2_META=""
+    for _i in $(seq 1 60); do
+        POST_NATIVE_META=$(pgq "SELECT encode(metadata,'hex') FROM faucet_registry WHERE lower(faucet_id)=lower('$CUSTOM_FAUCET_ID');")
```
mandrigin: **Restore identity coverage:** this proves only that a row with the faucet ID and matching metadata bytes exists. It does not prove it was recovered as the native route. Please assert the recovered native row's `origin_network == $MIDEN_NETWORK_ID`, `origin_address == $CUSTOM_ORIGIN`, symbol, and decimals; make the analogous origin/symbol/decimals assertions for the network-2 row. That is the explicit native + net-2 registry-row acceptance criterion.

mandrigin (commented):
Re-review of `88a3a7c..40bea59`: the FaucetKind guard is correct, both invalid BYTEA comparisons are fixed, all three mismatch cases now check for a row, and the restore flag has a real caller. The production change looks good. Remaining blockers are test evidence and destructive-suite coordination; the new legacy procedure should also be removed because the confirmed rollout is clean-slate. Static diff and shell syntax checks pass; CI is green with the opt-in L2L2 job skipped.

scripts/e2e-miden-origin.sh
```diff
@@ -126,6 +126,116 @@ done
...
+  \"params\":[{\"faucet_id\":\"$NATIVE_FAUCET_ID\",\"origin_token_address\":\"$BAD_NAME_ORIGIN\",
+    \"symbol\":\"MDN\",\"decimals\":8,\"name\":\"Not The Real Name\"}]}" 2>/dev/null)
+echo "$RESP" | grep -qi 'name mismatch' \
+  || fail "#149: wrong-name registration should error 'name mismatch', got: $RESP"
+_no_registry_row "$BAD_NAME_ORIGIN" || fail "#149: wrong-name registration must leave NO registry row"
+# No-bridge-note EVIDENCE (PR #150 re-review): register_faucet_in_bridge (the
+# ConfigAggBridgeNote) runs only AFTER validation passes. To prove the ON-CHAIN
+# config/token map has no entry for these origins (not merely that the direct
+# store-write was skipped), wait one full faucet-reconciler cycle (default 30s
+# poll) and re-assert NO row: the reconciler reads the bridge's on-chain config
```
mandrigin: **Blocking false-green:** the reconciler cannot provide the claimed on-chain evidence. `FaucetRegistryReconciler` is explicitly a tripwire, not an adopter: it enumerates faucet IDs and only checks `store.get_faucet_by_id(id)`. Every mismatch here reuses the already-registered `NATIVE_FAUCET_ID`, so even if a bug emitted a config note that associated that faucet with a bad origin, the reconciler would see the ID as known, materialize no bad-origin row, and all three checks would pass. Please inspect the bridge's on-chain origin/config mapping directly before/after each rejection, or add an endpoint-path test where a successful authoritative metadata read followed by a mismatch proves the bridge-registration client call is never made. The 35-second DB-row inference and its comments should be removed.

scripts/e2e-cantina13-metadata-recovery.sh
```diff
@@ -430,6 +430,27 @@ docker exec "$PG_CONTAINER" pg_dump -U agglayer agglayer_store > "$BACKUP" 2>/de
...
+# Before the from-scratch DROP + --restore, snapshot a native CUSTOM-NAME row and a
+# network-2 row so we can prove they SURVIVE recovery with FULL identity
+# (faucet_id | origin_network | origin_address | symbol | origin_decimals | preimage).
+# The native custom-name (WMDN, name != symbol) row comes from e2e-miden-origin.sh's
+# custom-name faucet (step 1d); a net-2 (L2B-origin) row comes from the l2l2 tier. This
+# reuses rows that already exist in the full suite — no extra registration, no separate
+# proxy-only reset (PR #150 re-review). If either row is absent (cantina13 run outside
+# the full suite), SKIP — the assertion is only meaningful with the prior tiers present.
+_s149_row() { pg "SELECT faucet_id||'|'||origin_network||'|'||encode(origin_address,'hex')||'|'||symbol||'|'||origin_decimals||'|'||encode(metadata,'hex') FROM faucet_registry WHERE $1 ORDER BY faucet_id LIMIT 1"; }
+S149_NATIVE_ROW=$(_s149_row "symbol='WMDN'")
```
mandrigin: **Restore identity selector:** `symbol='WMDN'` alone can select a foreign-origin WMDN row on a warm/shared stack; the later byte-for-byte comparison would then prove survival of the wrong route. Constrain this capture to the native route too, e.g. `origin_network=$LOCAL_NETWORK_ID` (while retaining the custom-name preimage assertion).

scripts/e2e-cantina13-metadata-recovery.sh
```diff
@@ -430,6 +430,27 @@ docker exec "$PG_CONTAINER" pg_dump -U agglayer agglayer_store > "$BACKUP" 2>/de
...
+S149_NET2_ROW=$(_s149_row "origin_network=2")
+if [[ -n "$S149_NATIVE_ROW" && -n "$S149_NET2_ROW" ]]; then
```
mandrigin, line 457: **Blocking acceptance false-skip:** in the normal full-overlay suite, the preceding L2L2/native phases are supposed to create both prerequisite rows, but the runner does not tell Cantina13 that they are mandatory. A regression in either phase therefore sets `S149_CHECK=0` and lets the full release gate pass without exercising #149 restore-survival. Keep the skip for standalone Cantina13 if useful, but have the full-overlay runner set a small flag such as `REQUIRE_S149_RESTORE=1`, and fail here when that flag is set and either row is absent.

mandrigin (commented):
Re-review of `48959f6`: moving #149 restore-survival into Cantina13 is the right architecture because it reuses the existing coordinated proxy + bridge-service recovery. The runbook, exact-zero/error-safe DB assertion, restore ordering, and full-row post-restore comparison are addressed. Three test-proof gaps remain below; I found no new production-code defect in this delta.

scripts/e2e-test.sh
mandrigin: Addressed in `48959f6`: the separate proxy-only destructive run was removed and the proof is folded into Cantina13's existing coordinated proxy + bridge-service recovery. This avoids the cached-nonce wedge and reuses already-created rows. Resolving; the full-suite prerequisite must-fail behavior is tracked on the current lines.

docs/operations/runbook.md
mandrigin: Addressed in `48959f6`: the runbook now explicitly forbids surgical row deletion/re-registration, preserves and quarantines unexpected legacy state, and limits the supported rollout to a clean rebuild. Resolving.

scripts/e2e-miden-origin.sh
mandrigin: Addressed in `48959f6`: `_no_registry_row` now propagates `pgq` failure and requires `COUNT(*)` to equal exactly `0`, so empty/error output cannot pass. Resolving.

scripts/e2e-miden-origin.sh
mandrigin: Addressed in `48959f6`: the pre/post snapshot compares faucet ID, origin network, origin address, symbol, decimals, and metadata bytes for both rows, with an explicit custom-name preimage assertion. Resolving; the native pre-capture selector itself is tracked separately on the current line.

scripts/e2e-miden-origin.sh
mandrigin: The latest attempt is still not sufficient, but this comment is outdated. Superseded by a current-line blocker on the 35-second reconciler inference: the reconciler is a tripwire, not an adopter, and the reused faucet ID is already known.

scripts/e2e-cantina13-metadata-recovery.sh
mandrigin: One more field-level gap in the same snapshot: `FaucetEntry` also contains `miden_decimals` and `scale`, but `_s149_row` omits both while the test claims “FULL identity.” A restore with incorrect conversion parameters could therefore pass. Please include both columns in the pre/post byte-for-byte tuple (and keep the expected native/net-2 route constraints).

scripts/e2e-miden-origin.sh
```diff
@@ -126,6 +126,110 @@ done
...
+_no_registry_row "$BAD_DEC_ORIGIN" || fail "#149: wrong-decimals registration must leave NO registry row"
+# wrong (non-matching) explicit name → 'name mismatch', and NO registry row
+RESP=$(curl -s "$L2_RPC" -H "Content-Type: application/json" -H "$ADMIN_BEARER" -d "{
+  \"jsonrpc\":\"2.0\",\"id\":1,\"method\":\"admin_registerNativeFaucet\",
+  \"params\":[{\"faucet_id\":\"$NATIVE_FAUCET_ID\",\"origin_token_address\":\"$BAD_NAME_ORIGIN\",
+    \"symbol\":\"MDN\",\"decimals\":8,\"name\":\"Not The Real Name\"}]}" 2>/dev/null)
+echo "$RESP" | grep -qi 'name mismatch' \
+  || fail "#149: wrong-name registration should error 'name mismatch', got: $RESP"
+_no_registry_row "$BAD_NAME_ORIGIN" || fail "#149: wrong-name registration must leave NO registry row"
+# The no-row checks above are the store-side evidence. That the bridge
```
mandrigin, line 175: **Blocking acceptance coverage remains:** this accurately describes current source ordering, but it is not an executable regression proving no `ConfigAggBridgeNote` was emitted. The cited mismatch tests call only the pure resolver; the E2E observes only the error and PostgreSQL. The endpoint call-count test covers unreadable metadata, not a successful authoritative read followed by a mismatch. Issue #149 explicitly requires a regression proving failed registration emits no bridge configuration. Please add the endpoint-path mismatch test (successful read, specific mismatch, second `miden_client.with()` not reached/call count remains one, no row), or directly compare authoritative bridge state before/after.

scripts/e2e-cantina13-metadata-recovery.sh
```diff
@@ -430,6 +430,37 @@ docker exec "$PG_CONTAINER" pg_dump -U agglayer agglayer_store > "$BACKUP" 2>/de
...
+# the full suite), SKIP — the assertion is only meaningful with the prior tiers present.
+# FULL identity: faucet_id | origin_network | origin_address | symbol |
+# origin_decimals | miden_decimals | scale | metadata-preimage — miden_decimals
+# and scale included so a restore with wrong CONVERSION params can't pass (PR #150).
+_s149_row() { pg "SELECT faucet_id||'|'||origin_network||'|'||encode(origin_address,'hex')||'|'||symbol||'|'||origin_decimals||'|'||miden_decimals||'|'||scale||'|'||encode(metadata,'hex') FROM faucet_registry WHERE $1 ORDER BY faucet_id LIMIT 1"; }
+# Constrain the native custom-name capture to the NATIVE route (origin_network ==
+# the proxy's own net id): symbol='WMDN' alone could select a foreign-origin WMDN
+# row on a warm/shared stack and prove survival of the WRONG route (PR #150).
+S149_NATIVE_ROW=$(_s149_row "symbol='WMDN' AND origin_network=${LOCAL_NETWORK_ID}")
+S149_NET2_ROW=$(_s149_row "origin_network=2")
```
mandrigin: **Blocking net-2 recovery false-green:** `_s149_row` returns a non-empty tuple even when its final `metadata` field is empty, so this prerequisite passes and the later equality accepts `empty -> empty`. That would claim a byte-identical network-2 metadata preimage without exercising recovery from its own network RPC. Require the selected net-2 row's metadata field to be non-empty before the drop (ideally select the known OPT0/L2B route), then retain the post-restore equality. The former dedicated leg explicitly required `PRE_NET2_META` to be non-empty.

mandrigin (commented):
Re-review of `79541a3` against `48959f6`: the mandatory full-suite restore gate and complete native/net-2 identity comparison are addressed, and the invalid reconciler inference is removed. The present production control flow is ordered safely. Two small but real acceptance false-greens remain below: executable proof that mismatch rejection skips bridge registration, and requiring the selected net-2 recovery preimage to be non-empty.

scripts/e2e-miden-origin.sh
mandrigin: Partially addressed in `79541a3`: the invalid reconciler inference and 35-second sleep are correctly removed, and current production ordering is safe. The required executable no-Config-note regression is still absent, so this outdated thread is superseded by a current-line comment explaining the remaining gap.

scripts/e2e-cantina13-metadata-recovery.sh
mandrigin: Addressed in `79541a3`: the full-overlay runner now sets `REQUIRE_S149_RESTORE=1`, and Cantina13 fails when either prerequisite row is absent in that mode while retaining the standalone skip. Resolving.

scripts/e2e-cantina13-metadata-recovery.sh
mandrigin: Addressed in `79541a3`: the native capture is constrained to `origin_network=$LOCAL_NETWORK_ID`, and the pre/post tuple now compares every `FaucetEntry` identity/conversion field, including `miden_decimals` and `scale`, plus metadata bytes. Resolving.

scripts/e2e-miden-origin.sh
mandrigin: Addressed in `03f05f7`: production now delegates the exact validate-then-register tail to `register_native_validated`. The new regression drives symbol, decimals, and name mismatches through that same tail, requires zero bridge-client calls and no row, then uses a matching positive control requiring one bridge-client call. This is executable evidence that validation gates `register_faucet_in_bridge`; resolving.

scripts/e2e-cantina13-metadata-recovery.sh
mandrigin: Addressed in `03f05f7`: both native and network-2 PRE selectors now require `octet_length(metadata) > 0`; under the required full-suite mode an empty preimage therefore fails before the destructive restore. The unchanged post tuple equality catches a non-empty preimage becoming empty. Resolving.

mandrigin (commented):
Re-review of `03f05f7` against `79541a3`: both remaining findings are addressed. The private `register_native_validated` seam is the exact production validate-then-register tail; mismatch cases require zero bridge-client calls/no row, while a matching positive control requires one call, so the regression is non-vacuous. The endpoint performs no mutation between its authoritative read and delegation to this seam. Native/network-2 restore selection now requires non-empty metadata and post-restore full-tuple equality detects loss. No new delta blocker found; static checks and CI are green.

IvanBelyakoff (approved):

PR #151, opened 2026-07-21
fix(recovery): gate readiness until historical claim calldata repair completes

### LINKED ISSUE #148: fix(recovery): gate readiness until historical claim calldata is repaired
## Context

Follow-up from [PR #145](https://github.com/gateway-fm/miden-agglayer/pull/145). This is intentionally separate from the release patch.

The supported full-disaster-recovery path rebuilds both proxy PostgreSQL state and the Miden client store before bridge-service consumers resume. In that path, node replay reconstructs the transaction envelope before publishing the corresponding synthetic `ClaimEvent`.

There is a conditional gap in a different recovery shape:

1. PostgreSQL is retained with a processed claim and `ClaimEvent`, but the stored transaction envelope/calldata is missing.
2. The Miden client store or reconciliation cursor is reset.
3. Bridge-service/API consumers reconnect before genesis reconciliation and historical calldata repair finish.
4. Restore skips the already-processed claim, while readiness can become green before the deferred repair completes.

During that window, `eth_getTransactionByHash` can expose an old claim transaction with empty input even though its `ClaimEvent` is already visible.

## Required decision

Explicitly decide and document whether **retained PostgreSQL + reset Miden store** is a supported recovery mode.

## Minimal scope

If supported:

- Keep readiness false until reconciliation reaches the current Miden tip and the historical claim-calldata repair pass completes successfully.
- Release bridge-service/API consumers only after the repaired transaction envelope is queryable.

If unsupported:

- Reject the flag/state combination with an actionable error, or require the runbook to keep bridge-service stopped and resync it only after repair completes.

Do not redesign the normal full-recovery path or introduce distributed-replica coordination here.

## Acceptance criteria

- A retained-PostgreSQL fixture starts with a historical `ClaimEvent` whose transaction envelope/calldata is missing.
- Readiness remains false until reconciliation reaches the Miden tip and repairs that envelope.
- Before consumers are released, `eth_getTransactionByHash` returns the complete original `claimAsset` calldata—not `0x` or reconstructed placeholders.
- Aggsender/bridge-service is then resynced and successfully settles the recovered claim/certificate.
- ClaimEvents originating from foreign AggLayer deployments remain absent throughout the same recovery test.
- The supported/unsupported status of this recovery shape is documented in the operator runbook.

## Non-goals

- Active-active/shared-PostgreSQL proxy replicas; tracked separately in [#142](https://github.com/gateway-fm/miden-agglayer/issues/142).
- Changes to ordinary full PostgreSQL + Miden-store recovery.

mandrigin:
## Summary

Supports retained-PostgreSQL + reset-Miden-store recovery by keeping the proxy unready while historical `ClaimEvent` transactions are missing their persisted `claimAsset` calldata.

## What changes

- tracks a durable backlog of claims whose transaction envelope is absent, pending, or failed;
- makes `/health` return `503` until the backlog is repaired and fails readiness closed if the query fails;
- re-observes historical CLAIM notes after a Miden-store/cursor reset and atomically replaces pending or failed envelopes with successful calldata;
- keeps AggKit, bridge-service, and autoclaim stopped during repair, then force-recreates consumers only after `/health` returns ready;
- changes the Compose healthcheck to probe `/health`;
- adds memory/PostgreSQL unit regressions and a destructive retained-Postgres recovery E2E;
- adds a dedicated Make/workflow invocation intended to exercise recovery three times outside `e2e-test.sh all`.

## Recovery flow

```mermaid
flowchart LR
    A[Retained proxy PostgreSQL] --> B[Historical ClaimEvent remains]
    C[Reset Miden store and reconciliation cursor] --> D[Genesis reconciliation]
    B --> E{Successful calldata envelope present?}
    E -- No --> F[/health: 503; consumers withheld]
    D --> G[Re-observe consumed CLAIM note]
    G --> H[Rebuild and atomically persist original claimAsset calldata]
    H --> E
    E -- Yes --> I[/health: 200 ready]
    I --> J[Force-recreate and resync consumers]
```

## Validation

- readiness/backlog transition and failure-path unit coverage;
- `scripts/e2e-recovery-readiness.sh` plus the dedicated `make e2e-recovery-readiness` and opt-in L2L2 workflow path;
- normal lint/unit `check`, CodeQL, and security checks green;
- `git diff --check` and shell syntax validation clean;
- opt-in L2L2 E2E was not selected for this push.

Closes #148


scripts/e2e-recovery-readiness.sh
```diff
@@ -0,0 +1,172 @@
...
+source "$SCRIPT_DIR/lib-l2l2.sh"
+
+CLAIM_EVENT_TOPIC="0x1df3f2a973a00d6635911755c260704e95e8a5876997546798770f76396fda4d"
+E2E_COMPOSE=(docker compose -f "$PROJECT_DIR/docker-compose.e2e.yml" -f "$PROJECT_DIR/docker-compose.l2l2.yml" --env-file "$FIXTURES_DIR/.env")
+PG_CONTAINER="${PG_CONTAINER:-${COMPOSE_PROJECT_NAME}-agglayer-postgres-1}"
+BRIDGE_PG_CONTAINER="${BRIDGE_PG_CONTAINER:-${COMPOSE_PROJECT_NAME}-postgres-1}"
+
+# `pgi` — a COUNT/scalar query that FAILS on a psql error (no false-green from a
+# suppressed error; PR-review-rigor #1). `pg` (from lib-l2l2) is used for text.
+pgi() { local out; out=$(pg "$1") || return 1; printf '%s' "$out"; }
```
mandrigin: **Blocking:** `lib-l2l2.sh` defines `pgq`, not `pg`, so this helper and the direct calls below invoke a nonexistent command. With the current `|| true` around claim selection, standalone mode silently turns that into a skip; mandatory mode fails before testing recovery. Please use `pgq` consistently (including the direct update/query calls) and keep mandatory mode non-skippable.

scripts/e2e-recovery-readiness.sh
```diff
@@ -0,0 +1,172 @@
...
+# RETAIN Postgres; only remove the claim's calldata envelope (transaction_logs
+# cascade-drops via FK; the ClaimEvent in synthetic_logs is UNTOUCHED), and reset
+# the reconcile cursor to 0 so the genesis re-sweep re-observes + backfills it.
+DELETED="$(pgi "WITH d AS (DELETE FROM transactions WHERE lower(tx_hash) = '$CLAIM_TX' RETURNING 1) SELECT COUNT(*) FROM d")"
+[[ "$DELETED" == "1" ]] || fail "#148: expected to blank exactly 1 claim tx envelope, deleted '$DELETED'"
+pg "UPDATE service_state SET reconcile_cursor = 0 WHERE id = 1" >/dev/null
+# Reset the Miden client store (the issue's recovery shape) — best-effort delete
+# of the sqlite from the bind mount; the reconcile_cursor=0 above independently
+# forces the re-sweep, so the repair runs whether or not the rm matched.
+rm -f "$PROJECT_DIR/.miden-agglayer-data/store.sqlite3"* 2>/dev/null || true
```
mandrigin: **Blocking:** the defining recovery precondition is explicitly best-effort here. `rm -f ... || true` lets this test pass with the original Miden sqlite intact, exercising only a Postgres cursor reset rather than retained-Postgres + reset-Miden-store. Please use the real `--reset-miden-store` path (or fail on deletion and assert the old sqlite is absent / a fresh one is created) so a passing test proves the supported operator recovery shape.

scripts/e2e-recovery-readiness.sh
```diff
@@ -0,0 +1,172 @@
...
+# Realistic operational resync (finding #65): the reset proxy nonces are 0, so
+# resync the bridge-service alongside so it re-fetches nonce 0 (else future-nonce
+# wedge). Then assert it comes back reachable + synced against the recovered proxy.
+"${E2E_COMPOSE[@]}" stop bridge-service bridge-autoclaim >/dev/null 2>&1
+docker exec "$BRIDGE_PG_CONTAINER" psql -U bridge_user -d bridge_db \
+    -c "DROP SCHEMA IF EXISTS sync CASCADE; DROP SCHEMA IF EXISTS mt CASCADE; DROP SCHEMA public CASCADE; CREATE SCHEMA public; GRANT ALL ON SCHEMA public TO bridge_user;" >/dev/null 2>&1 \
+    || fail "#148: failed to drop bridge_db for the realistic resync (finding #65)"
+"${E2E_COMPOSE[@]}" up -d --no-deps bridge-service bridge-autoclaim >/dev/null 2>&1
+wait_for "bridge-service resynced + reachable after recovery" \
+    "[[ \$(curl -s -m3 -o /dev/null -w '%{http_code}' $BRIDGE_SERVICE_URL/ 2>/dev/null) =~ ^(200|404)\$ ]]" 180 5
```
mandrigin: **Blocking acceptance gap:** this proves only that the bridge-service process answers 200/404. It remains running while the proxy recovers (Compose `depends_on` does not stop an existing consumer, and the proxy container healthcheck still probes `eth_chainId`, not `/health`), then its DB is reset and no recovered ClaimEvent indexing or certificate settlement is asserted. Stop/unpoint consumers before recovery, start them only after `/health` becomes 200, and assert actual sync/index progress plus successful settlement of the recovered claim/certificate. If Compose is the intended gate, its healthcheck must use `/health`.

src/store/postgres.rs
```diff
@@ -238,6 +238,28 @@ impl Store for PgStore {
...
+        // (idx_logs_tx_hash / transactions PK) and beats a NOT IN subquery.
+        let row = client
+            .query_one(
+                "SELECT COUNT(*) FROM (
+                     SELECT DISTINCT lower(transaction_hash) AS h
+                     FROM synthetic_logs
+                     WHERE lower(topics[1]) = lower($1)
+                 ) c
+                 LEFT JOIN transactions t ON lower(t.tx_hash) = c.h
+                 WHERE t.tx_hash IS NULL",
```
mandrigin: **Blocking:** any `transactions` row clears the backlog, regardless of `status`. Repair persists via `txn_begin` and then a separate fallible `txn_commit`; between those calls—or permanently after a crash/commit failure—`/health` can report ready. The receipt path treats that known pending row as `null` and suppresses the synthetic-log fallback, so the consumer can still stall. Count absent **or non-success** rows (or make repair insertion/finalization atomic) and add a crash-after-`txn_begin`/restart regression.

src/store/postgres.rs
```diff
@@ -238,6 +238,28 @@ impl Store for PgStore {
...
+        // Distinct ClaimEvent tx-hashes (topic0 == CLAIM_EVENT_TOPIC) that have NO
+        // row in `transactions` — i.e. their `claimAsset` calldata envelope was
+        // never persisted or was lost. LEFT JOIN + IS NULL is indexable
+        // (idx_logs_tx_hash / transactions PK) and beats a NOT IN subquery.
+        let row = client
+            .query_one(
+                "SELECT COUNT(*) FROM (
+                     SELECT DISTINCT lower(transaction_hash) AS h
+                     FROM synthetic_logs
+                     WHERE lower(topics[1]) = lower($1)
```
mandrigin: **Production concern:** this is not the indexed scan the comments claim. There is no expression/partial index for `lower(topics[1])`, and `lower(t.tx_hash)` cannot use the plain `transactions` primary key. Every readiness poll therefore scans/deduplicates historical logs and performs a lowercased join. Please use normalized exact hashes plus an appropriate topic/partial index, or maintain a cheap durable recovery state/counter instead of doing unbounded historical work on `/health`.

src/store/memory.rs
```diff
@@ -445,6 +445,35 @@ impl Store for InMemoryStore {
...
+            let has_calldata = tx_hash
+                .parse::<TxHash>()
+                .ok()
+                .is_some_and(|h| txns.contains(&h));
+            if !has_calldata {
+                awaiting += 1;
+            }
+        }
+        Ok(awaiting)
+    }
```
mandrigin, line 500: The LRU note is backwards: eviction from the 10,000-entry `transactions` cache while `logs_by_tx` remains populated makes this **overcount**, eventually putting an ordinary long-running in-memory service into permanent recovery/503. Either leave the retained-Postgres gate disabled for this non-durable backend or track claim-envelope completeness outside the evicting transaction cache.

mandrigin (commented):
Review result: **not ready yet**. The recovery direction is sound and maps directly to #148, but the current gate and acceptance test have correctness holes that can release consumers while the receipt is still unusable, or prevent the mandatory E2E from exercising the path at all. I have left the concrete blockers inline. Shell syntax and `git diff --check` pass; CI is still running.

Non-blocking documentation follow-up within this PR: `README.md`, `docs/operations/monitoring.md`, and `docs/operations/diagnostics.md` still describe `/health` as liveness-only / 503 only for node loss, which contradicts the new recovery semantics. Please update those references together with the runbook.

Posted as a comment review because the PR author and reviewer are the same account.

src/store/postgres.rs
mandrigin: The original false-ready condition is addressed in `0225654`: the durable backlog includes absent/non-success rows, and a successful PostgreSQL `txn_commit` removes the pending key in the same transaction as the success update. Pending/crash and atomic-drain regressions were added. Resolving this original thread; the distinct failed-terminal healing gap is called out separately.

scripts/e2e-recovery-readiness.sh
mandrigin: This is substantially improved, but **still blocking**. Stopping the consumers during repair is correct. However, `up -d` restarts the same `aggkit` container; `PathRWData=/tmp` is inside that preserved container, so AggKit resumes from its old bridgesync cursor and need not fetch or parse this historical claim again. The queried `sync.status` belongs to bridge-service and proves its chain scan caught up; it is not an AggKit exact-claim parse or an AggLayer certificate settlement.

Minimal acceptance fix: force-recreate AggKit (fresh `/tmp`), prove its L2 sync starts from zero and advances past this exact claim block, require a new proxy serve of this exact claim hash, bind the recovered claim by exact global index in bridge-service, then require a fresh post-reset certificate to reach `Settled` (ideally verify its settlement receipt hits RollupManager). `e2e-synthesized-claim-calldata.sh` already contains these non-vacuous gates and can be reused. Also make the Compose healthcheck probe `GET /health` if `depends_on: service_healthy` is meant to enforce the release gate; it currently probes `eth_chainId`.

scripts/e2e-recovery-readiness.sh
mandrigin: Addressed in `0225654`: all proxy-DB reads now use `pgq` through the error-propagating `pgi` helper, and mandatory mode still fails when no qualifying claim exists. Resolving.

scripts/e2e-recovery-readiness.sh
mandrigin: Addressed in `0225654`: the test removes the exact sqlite files as root through the shared bind mount, requires that at least one file existed, and asserts that none remains before restart. This now exercises the real reset-store shape. Resolving.

src/store/postgres.rs
mandrigin: Addressed in `0225654`: `/health` now counts the small durable pending set; the historical log/transaction reconciliation runs once at startup rather than on every readiness poll. Resolving.

src/store/memory.rs
mandrigin: Addressed in `0225654`: the memory backend now tracks repair state in a separate set, independent of the evicting transaction LRU, and includes an eviction regression proving a repaired claim cannot reappear in the backlog. Resolving.

src/store/postgres.rs
```diff
@@ -238,6 +238,69 @@ impl Store for PgStore {
...
+        // legacy non-canonical rows.
+        tx.execute(
+            "INSERT INTO claim_calldata_repair_pending (tx_hash)
+             SELECT c.h FROM (
+                 SELECT DISTINCT lower(transaction_hash) AS h
+                 FROM synthetic_logs
+                 WHERE lower(topics[1]) = lower($1)
+             ) c
+             LEFT JOIN transactions t ON lower(t.tx_hash) = c.h
+             WHERE t.tx_hash IS NULL OR t.status <> 'success'
```
mandrigin, line 278: **Blocking:** this seed correctly includes `status = 'failed'`, but the repair path treats every terminal result as complete: `persist_synthetic_claim_tx` returns `Ok(true)` for `Some(Err(_))`. Such a claim stays in `claim_calldata_repair_pending`, is never committed successfully, and leaves `/health` at 503 forever.

Minimal fix: only `Some(Ok(_))` is already complete; route both pending and failed rows through the existing success-overrides-failure `txn_commit(Ok(...))`, then add a failed-row → repair → backlog-zero regression for the production PostgreSQL path.

scripts/e2e-recovery-readiness.sh
```diff
@@ -0,0 +1,236 @@
...
+
+# ── 1. Readiness is WITHHELD while the claim calldata is missing ──────────────
+# Poll: we MUST observe a 503 "recovering" with claims_awaiting_calldata >= 1 at
+# least once (the gate holding), then a transition to 200 (repair complete).
+SAW_RECOVERING=0; READY=0
+for _i in $(seq 1 200); do
+    CODE="$(proxy_health_code)"
+    if [[ "$CODE" == "503" ]]; then
+        BODY="$(proxy_health_body)"
+        if echo "$BODY" | grep -q 'recovering' && echo "$BODY" | python3 -c "import json,sys;sys.exit(0 if (json.load(sys.stdin).get('claims_awaiting_calldata',0) or 0) >= 1 else 1)" 2>/dev/null; then
```
mandrigin: **Blocking E2E race:** the production client sets `is_alive = true` only *after* its initial sync listeners finish. On this reset path the same initial projector tick can catch the reconciler up and repair the only missing claim, so `/health` transitions from `degraded` (alive=false, backlog>0) directly to `200` (alive=true, backlog=0). A healthy implementation can therefore never satisfy this mandatory `SAW_RECOVERING` assertion on a small/fast chain.

Keep the assertion non-vacuous without depending on that presentation race: prove the durable pending count becomes `>= 1`, observe any 503 while it is pending, then require pending count `0`, reconcile cursor at the captured Miden tip, full calldata restored, and only then `/health=200`. Alternatively expose/prioritize the recovery backlog even while node liveness is still false.

mandrigin (commented):
Re-review of `0225654`: **materially improved, not ready yet**.

Five original findings are fully addressed and resolved: `pgq`/error propagation, real Miden-store reset, atomic durable backlog handling for pending rows, removal of the per-health historical scan, and memory-LRU independence.

Three blockers remain inline:

1. a terminal failed claim is counted but never healed, permanently holding `/health` at 503;
2. the mandatory E2E depends on observing a `recovering` state that can be skipped legitimately during the initial sync (`degraded` → ready);
3. the consumer test preserves AggKit's `/tmp` sync DB and checks bridge-service catch-up, not a fresh exact-claim reparse and certificate settlement.

Current CI is green; the L2L2 E2E check was skipped, so it does not validate these runtime paths.

scripts/e2e-recovery-readiness.sh
mandrigin: Addressed in `3fb1029`: degraded health responses now expose `claims_awaiting_calldata`, and the E2E asserts the actual invariant—HTTP 503 plus backlog ≥1—rather than requiring the transient `recovering` label. This covers the valid degraded→ready transition. Resolving.

src/store/postgres.rs
mandrigin: Addressed in `e98b134`: `persist_synthetic_claim_tx` now treats only `Some(Ok(_))` as complete and re-commits both pending and failed rows through the success-overrides-failure path, which also drains the durable backlog. The new end-to-end store regression covers the branch; the existing PgStore regressions separately pin failed→success and success→backlog drain. Resolving.

scripts/e2e-recovery-readiness.sh
mandrigin: Partially addressed in `e98b134`: AggKit is now force-recreated with a fresh `/tmp` cursor, and the script proves it advances past the recovered claim block. Keeping this open: it still does not prove a new serve of the exact recovered hash, bind the exact global index in AggKit/bridge-service, or require a fresh certificate to reach `Settled`; the Compose healthcheck also still probes `eth_chainId`, so `service_healthy` is not the new readiness gate. In addition, `5a4683d` removed this destructive test from `e2e-test.sh all`, but there is no Make target/workflow/other repository invocation for the claimed dedicated 3× gate.

mandrigin: Updated review at `8ccad40`: the force-recreated AggKit cursor, Compose `/health` gate, exact ClaimEvent↔calldata global-index check, and dedicated 3× invocation are good improvements.

**Keeping this acceptance thread blocking for three tightly related gaps:**

1. `FRESH_SERVE="$(tx_input "$CLAIM_TX")"` is the test script itself calling the proxy. It proves the repaired bytes, but not that force-recreated AggKit fetched this exact hash. Snapshot the proxy exact-hash serve count before consumer release and require a strict post-recreate increase attributable to AggKit.
2. The global index is bound inside the proxy only; require the same exact GI to appear as delivered in the consumer/bridge-service index.
3. No post-recovery outbound activity is created. The standalone Make target lands only L1→Miden (an imported claim, not a local-exit-tree leaf) and then requires a non-empty `NewLocalExitRoot`, so a fresh stack can legitimately time out. Conversely, a restart may re-log an old settled certificate. After recovery, submit one small Miden bridge-out, require a certificate newer than the pre-recovery cert (ID/height), and receipt-check its `SettlementTxnHash` against RollupManager.

`e2e-synthesized-claim-calldata.sh` already has the exact-hash count delta, exact-GI consumer check, post-reset bridge-out, and settlement evidence patterns. Reusing those is the minimal non-vacuous fix.

IvanBelyakoff (approved):

PR #152, opened 2026-07-21
test(e2e): verify received faucet identity and metadata across L1 and L2B

### LINKED ISSUE #147: test(e2e): verify faucet symbols for native L1/L2B and ERC-20 bridge-ins
## Problem

A wallet displayed an incoming AggLayer P2ID note as `Unknown` instead of `ETH`:

- [Sepolia deposit](https://sepolia.etherscan.io/tx/0x9097fe3270e358ff33159a4cf952a95b8c107a11cf9ab8406460a5e04a831bd4)
- [Proxy transaction hash shown in Midenscan](https://testnet.midenscan.com/tx/0x6a125266413e651c8a13a8cd223d468479d89dc9609fb8c32b9c9b5c2f3d27cf)

The Sepolia transaction is a successful native-ETH `bridgeAsset` deposit: origin network `0`, zero token address, empty metadata, destination network `78`, and amount `0.00192 ETH`. Empty bridge metadata is expected for native ETH. The configured public Miden faucet currently resolves to `symbol=ETH, decimals=8`, so the observed UI symptom is likely a wallet metadata-resolution issue rather than incorrect proxy metadata.

We nevertheless lack an E2E contract proving that every faucet produced or selected by the proxy exposes the correct wallet-resolvable token metadata.

## Current coverage gap

- [`e2e-l1-to-l2.sh`](https://github.com/gateway-fm/miden-agglayer/blob/main/scripts/e2e-l1-to-l2.sh) bridges native L1 ETH and verifies balance with a pre-known faucet ID; it does not resolve or assert that faucet's public symbol/decimals.
- [`e2e-l2l2-forward.sh`](https://github.com/gateway-fm/miden-agglayer/blob/main/scripts/e2e-l2l2-forward.sh) covers an L2B-origin ERC-20 (`OPT0`), but not native L2B ETH.
- [`e2e-dynamic-erc20.sh`](https://github.com/gateway-fm/miden-agglayer/blob/main/scripts/e2e-dynamic-erc20.sh) verifies registry identity and balance, but not metadata fetched from the public Miden faucet account.
- The isolated-wallet helper passes a known faucet ID directly, so all current tests can stay green while a fresh wallet renders the received asset as `Unknown`.

A fungible P2ID asset carries the faucet ID and amount, not a symbol string. A client must resolve the public faucet account's token configuration (for example with `Client::fetch_remote_token_metadata`) to obtain the display symbol and decimals.

## Minimal change

Add a small machine-readable faucet-inspection mode to the existing isolated-wallet/bridge-out tool. It must use the same fresh client store and Miden node as the receiving wallet, fetch the public faucet metadata by the asset's faucet ID, and return at least `faucet_id`, `symbol`, and `decimals`. Do not obtain the expected display metadata from `admin_listFaucets` or a hardcoded wallet token map.

Use it in the existing E2Es, adding only the missing native-L2B deposit leg.

## Required matrix and assertions

| Origin | Asset | Expected Miden metadata | Identity assertion |
|---|---|---|---|
| L1 / network `0` | native ETH (`address=0`, `metadata=0x`) | `ETH`, 8 decimals | canonical seeded L1 ETH faucet |
| L2B / configured L2B network | native ETH (`address=0`, `metadata=0x`) | `ETH`, 8 decimals | a distinct faucet keyed by `(L2B network, zero address)` |
| L1 | dynamic ERC-20 | sanitized on-chain symbol, `min(origin_decimals, 8)` | exact `(origin_network, origin_address)` |
| L2B | existing `OPT0` ERC-20 | `OPT0`, 8 decimals | exact `(origin_network, origin_address)` |

For every row:

1. Start from an isolated receiving wallet with no preloaded symbol map.
2. Complete deposit -> claim -> MINT -> P2ID receipt/consumption.
3. Derive the faucet ID from the received fungible asset and fetch that faucet's public token metadata from Miden.
4. Assert the symbol and decimals above; `None`, RPC failure, or `Unknown` must fail the test.
5. Assert the wallet vault balance is credited under that same faucet ID.

The two native-ETH faucets must not collapse into one registry identity: `(0, 0x0)` and `(L2B_NETWORK_ID, 0x0)` represent different origins even though both display `ETH`.

## Acceptance criteria

- Native ETH is exercised from both L1 and L2B end-to-end.
- A fresh Miden client resolves `ETH/8` for both native-origin faucets from public account state.
- The faucets are distinct and correctly keyed by origin network.
- Existing L1 and L2B ERC-20 E2Es also assert remotely resolved symbol/decimals.
- Tests run in the normal CI E2E suite and print faucet ID, origin tuple, note ID, and fetched metadata on failure.

## Hash/explorer note

The supplied `0x6a1252...27cf` value is the proxy's EVM-compatible transaction hash, not necessarily the underlying Miden `TransactionId`; a Midenscan `transaction not found` result for that value does not show that the bridge claim failed. Keep the two identifiers distinct in test diagnostics.

mandrigin:
## Summary

Adds end-to-end coverage proving that a receiving Miden wallet discovers the faucet belonging to the asset it actually received, then resolves that faucet's public symbol and decimals without a preloaded token map.

This covers native and ERC-20 bridge-ins from both L1 and L2B, including the sovereign chain's native custom gas token.

## What changes

- adds `bridge-out-tool --inspect-faucet`, backed by `Client::fetch_remote_token_metadata`;
- adds `bridge-out-tool --list-wallet-faucets`, which syncs the isolated receiver, consumes pending P2ID notes, and reports resulting vault balances;
- snapshots the receiver vault before each deposit and derives the received faucet from the per-faucet balance delta;
- asserts that the derived faucet ID equals the expected origin-derived ID;
- resolves symbol and decimals for that same derived ID with a fresh temporary client store;
- retains explicit failures for missing metadata, RPC errors, wrong faucet identity, balance, symbol, or decimals.

## Coverage

| Source | Asset | Received-wallet assertion |
|---|---|---|
| L1 | native ETH | derived faucet ID + balance + `ETH / 8` |
| L1 | dynamic ERC-20 | derived faucet ID + balance + sanitized symbol / `min(decimals, 8)` |
| L2B | OPT0 ERC-20 | derived faucet ID + balance + `OPT / 8` |
| L2B | native custom gas token | derived distinct faucet ID + balance + `LBGAS / 8` |

## Native L2B architecture and recovery

The native-L2B leg resolves internal test finding 77. L2B is a custom-gas-token sovereign chain: its genesis-minted native balance is not backed by the bridge's Local Balance Tree, so the first native out-bridge previously reverted with `LocalBalanceTreeUnderflow`.

The test now:

1. deploys/configures `L2BGAS` as the L2B bridge gas token;
2. lazily seeds its Local Balance Tree by bridging and claiming the gas token from L1 to L2B;
3. bridges L2B native value out with `address(0)` and claims it on Miden;
4. verifies the resulting faucet uses the gas token's actual origin tuple `(0, L2BGAS address)`, rather than incorrectly treating it as `(L2B, address(0))`;
5. derives that faucet from the receiving wallet's vault delta and verifies public metadata from a cold client;
6. asserts the Miden-compatible sanitized symbol `L2BGAS -> LBGAS`, decimals `18 -> 8`, a distinct identity from L1 ETH, and a landed `ClaimEvent`.

The expensive native leg remains selectable with `RUN_L2B_NATIVE_ETH=1` for the full L2L2 run; it is a working, fresh-stack-validated path rather than a placeholder.

## Validation

- native L2B custom-gas-token leg validated end to end on a fresh stack;
- normal lint/unit `check` workflow green;
- CodeQL and security checks green;
- shell syntax validation for changed scripts;
- `git diff --check` clean.

Closes #147


scripts/e2e-l2l2-forward.sh
```diff
@@ -200,6 +207,66 @@ evidence_record "leg2b" forward Miden claim "" "${FWD_CLAIM_BLOCK:-}" "$BRIDGE_I
...
+# 0x14603c01 — confirmed via cast estimate (the pre-submit path), well-formed
+# call, ADMIN funded with ~9999 ETH on anvil-l2b. Bridging the sovereign chain's
+# native gas token OUT is not supported in this setup; the ERC-20 path (leg 2 /
+# OPT0) works fine. So this leg (and its "two native-ETH faucets distinct by
+# origin (0,0x0) vs (L2B_NET,0x0)" distinctness check) cannot be exercised until
+# sovereign native-ETH bridging (or a WETH-mapped path) is enabled. #147's core
+# wallet-resolvable-metadata coverage is proven by rows 1/3/4 (L1 native ETH, L1
+# ERC-20, L2B OPT0). When re-enabled, a bridgeAsset revert must fail LOUD with
+# this context, not a cryptic "bridgeAsset failed".
+if [[ "${RUN_L2B_NATIVE_ETH:-0}" == "1" ]]; then
```
mandrigin, line 241: **Blocking / KISS:** this entire leg is disabled by default and known to be unexecutable with the current sovereign bridge, but it also contains untested defects: deposit lookup strips `0x` even though `find_deposit` exact-compares bridge-service addresses; an empty `transactionHash` still reaches PASS; and `ETH_MIDEN_UNITS` is never checked against a consumed P2ID/wallet balance. The minimal honest change is to remove this dead block and leave native-L2B coverage in #147 until the bridge supports it. If it is retained, fix all three and wire it into an actual executable test before merging.

scripts/lib-isolated-wallet.sh
```diff
@@ -86,6 +86,44 @@ iso_wallet_balance() {
...
+# exactly what a cold receiving wallet does to render an incoming fungible P2ID
+# asset (#147). Echoes the tool's `inspect-faucet: faucet_id=.. symbol=.. decimals=..`
+# line on success (rc 0); returns NON-ZERO (echoing the ERROR line) when the metadata
+# is unresolvable (the wallet's `Unknown`) or the RPC fails — never a silent empty.
+# A throwaway store dir on the same bind-mount device is created + wiped each call.
+iso_inspect_faucet() {
+    local fid="$1" fresh rc out
+    fresh="$B2AGG_STORE_DIR/inspect-$$-${RANDOM}"
+    mkdir -p "$fresh/tmp"
+    out=$(docker run --rm --network "$ISO_NETWORK" \
```
mandrigin: Non-blocking robustness: when `iso_inspect_faucet` is invoked directly under `set -e`, a failed Docker assignment can exit before `rc` capture, cleanup, and the useful diagnostic. Current `assert_faucet_symbol ... || fail` call sites happen to suppress errexit for the function body. Please make the helper self-contained with `if out=$(docker run ...); then rc=0; else rc=$?; fi`.

scripts/lib-isolated-wallet.sh
```diff
@@ -86,6 +86,44 @@ iso_wallet_balance() {
...
+    return "$rc"
+}
+
+# assert_faucet_symbol <faucet_id> <expected_symbol> <expected_decimals> <origin-label>
+# → inspect the faucet on a fresh client and fail (with full diagnostics) unless it
+# resolves to exactly <expected_symbol>/<expected_decimals>. #147: a fresh wallet must
+# NOT render this asset as `Unknown`. Exports INSPECT_SYMBOL/INSPECT_DECIMALS.
+assert_faucet_symbol() {
+    local fid="$1" want_sym="$2" want_dec="$3" label="$4" insp
+    insp=$(iso_inspect_faucet "$fid") \
```
mandrigin, line 119: Acceptance scope note: this proves a caller-supplied faucet ID is remotely resolvable, which is a valuable proxy-side guard, but it does not yet prove the receiving wallet derives that ID from the incoming fungible P2ID asset. The current tests obtain IDs from `bridge_accounts.toml`, admin RPC, or Postgres and inspect them in a separate metadata-only store. Keep #147 open; completion should expose the received asset's faucet ID/note ID from the isolated wallet and resolve that same ID through the receiving-wallet path.

mandrigin (commented):
Review result: **useful partial coverage, but not ready as pushed**. The enabled `--inspect-faucet` path is sound: it uses miden-client's direct public-account lookup from an empty store, and the existing L1 ETH, dynamic ERC-20, and L2B ERC-20 assertions bind an explicit faucet ID to both balance and remotely fetched metadata. `cargo check --bin bridge-out-tool`, shell syntax checks, and `git diff --check` pass.

The default-off native-L2B block should not merge in its current broken/unexecutable form; the simplest KISS fix is to remove it and keep that matrix row tracked in #147 until the sovereign bridge has a supported native/WETH exit path. This PR must remain a partial `Refs #147`, not close the issue. Full #147 completion also still needs the faucet ID derived from the received asset, the same receiving-wallet resolution path, and note-ID diagnostics.

Copilot was requested but could not review because the requester quota is exhausted. Posted as a comment review because the PR author and reviewer are the same account.

scripts/e2e-l2l2-forward.sh
mandrigin: Scope update: native L2B is explicitly owned by a follow-up PR, so it is not a merge requirement for #152. I am downgrading and resolving this thread here. The env-gated path remains off; before the follow-up enables it, it must fix address normalization, require a real transaction hash/receipt, and verify P2ID consumption plus the expected wallet balance.

scripts/lib-isolated-wallet.sh
mandrigin: Confirmed as follow-up scope: #152 provides the public-account discoverability guard; deriving the faucet ID/note ID from the received asset and exercising the same receiving-wallet resolver will be completed separately. Resolving this scope thread for #152.

mandrigin (commented):
Scope update after product decision: native L2B coverage is owned by a follow-up PR and is not a blocker for #152. I resolved the two deferred-scope threads and corrected the PR body so it references—rather than closes—#147. No blocking finding remains for the enabled #152 behavior; the direct-call cleanup comment in lib-isolated-wallet.sh is non-blocking.

scripts/lib-isolated-wallet.sh
mandrigin: Correction after scope clarification: **this remains blocking for #152 / #147**. Only the native-L2B case is deferred.

The minimal missing proof is the received-asset linkage, not another resolver architecture: after the existing isolated receiving wallet syncs and consumes the P2ID note, capture that consumed note ID, derive the faucet ID from its fungible asset (or an unambiguous vault delta), fetch public metadata for that **derived** ID, and assert the expected symbol/decimals plus the wallet balance under the same ID. On failure, print the note ID, derived faucet ID, origin tuple, and fetched metadata.

Until that path is exercised, the test can pass for a correctly registered caller-supplied faucet ID even if the wallet cannot discover the asset it actually received. The existing known-ID resolver check is useful and should stay; native-L2B coverage remains non-blocking here.

mandrigin: Still blocking at `3237832`. `assert_received_faucet` accepts a caller-supplied faucet ID, polls balance for that same known ID, and explicitly defers enumerating the consumed note/asset. It is also called only by the default-off native-L2B leg; the three enabled rows still call `assert_faucet_symbol` with IDs sourced from config/admin RPC/Postgres. A retained balance can therefore false-pass without linking this deposit. The minimal completion remains: expose the consumed P2ID note ID and its fungible faucet ID (or an unambiguous before/after vault delta) from the receiving wallet, then use that derived ID for balance + cold metadata assertions in all three enabled rows, with note/faucet/origin/metadata diagnostics.

scripts/lib-isolated-wallet.sh
mandrigin: Addressed in `e3f9c45`: the Docker assignment is now wrapped in an explicit `if`, so `set -e` cannot bypass exit-code capture, cleanup, or diagnostics. Resolving.

scripts/lib-isolated-wallet.sh
mandrigin: Addressed in `07a5e7d`. Each of the three enabled rows now snapshots the isolated receiving wallet vault before the deposit, syncs and consumes its pending P2ID notes, derives the received faucet from the per-faucet balance delta, asserts that derived ID equals the expected origin faucet, and cold-fetches symbol/decimals for that same ID. A retained balance can no longer satisfy the assertion. Native L2B remains intentionally follow-up scope. Resolving.

mandrigin (commented):
Re-review at `07a5e7d`: **code-ready**. The receiving-wallet linkage is non-vacuous: before/after vault deltas derive the faucet actually received, the derived ID is matched to the origin faucet, and public metadata is resolved for that same ID using a cold client. This applies to L1 native ETH, L1 dynamic ERC-20, L2B OPT0, and the native L2B custom-gas-token leg implemented under internal finding 77. The native leg seeds the LBT, bridges and claims successfully, verifies a distinct faucet, and resolves `LBGAS/8`. Normal CI/security checks are green; the full native leg was validated on a fresh stack.

IvanBelyakoff (approved):

mandrigin (commented):
Correction after rechecking the internal finding 77 implementation: the native L2B custom-gas-token leg is already a real, fresh-stack-validated path in this PR. It seeds the Local Balance Tree, bridges native L2B value out, derives the received Miden faucet from the wallet delta, and cold-resolves the sanitized LBGAS/8 metadata. The missing per-note ID in failure output is a useful observability follow-up, but it does not invalidate the symbol/identity test and is not a merge blocker. With the native leg included, #152 completes the functional scope of #147.

PR #155, opened 2026-07-22
feat(rpc): persist and promote future-nonce transactions

### LINKED ISSUE #146: Queue future-nonce transactions like Geth instead of rejecting after 30s
## Problem

The proxy does not currently emulate Geth's handling of a valid transaction whose nonce is above the sender's next executable nonce.

Geth accepts a nonce-gapped transaction into its `queued` pool, returns the transaction hash, and promotes it once the missing nonce(s) arrive. Its `pending` nonce advances only over the contiguous executable prefix.

At reviewed head `5968603e2b155f47ebf89c5b28672e91db2a28d2`, the proxy instead:

1. holds the `eth_sendRawTransaction` request open while polling for at most 30 seconds;
2. returns `nonce mismatch` if the gap remains; and
3. does this before reservation, durable admission, or insertion into the writer queue.

The existing writer `mpsc` is therefore an execution queue **after** nonce admission, not an input/future-nonce queue.

Relevant code:

- [`service_send_raw_txn.rs` future-nonce wait/rejection](https://github.com/gateway-fm/miden-agglayer/blob/5968603e2b155f47ebf89c5b28672e91db2a28d2/src/service_send_raw_txn.rs#L1331-L1430)
- [`service_send_raw_txn.rs` durable admission and writer enqueue](https://github.com/gateway-fm/miden-agglayer/blob/5968603e2b155f47ebf89c5b28672e91db2a28d2/src/service_send_raw_txn.rs#L568-L609)
- [`writer_worker.rs` FIFO `mpsc`](https://github.com/gateway-fm/miden-agglayer/blob/5968603e2b155f47ebf89c5b28672e91db2a28d2/src/writer_worker.rs#L429-L473)

## Why this matters

This creates a conditional liveness loop with an otherwise normal Ethereum transaction manager:

```text
proxy expects nonce N
  -> an older retry is submitted first with nonce N+2
  -> proxy blocks that request waiting for N
  -> the sender's sequential monitor loop cannot submit N while blocked
  -> proxy rejects N+2 after 30 seconds
  -> the retry remains first and may repeat indefinitely
```

This is not a fund-safety failure—the proxy does not execute transactions out of order—but it can stop all progress for that signer until intervention. Under Geth, the N+2 submission returns immediately as queued, allowing the sender to continue and submit N.

## Expected behavior

Implement a small, bounded future-nonce reorder buffer suitable for this allow-listed proxy:

- Validate authentication, signature, chain ID, supported selector, and other deterministic admission rules.
- Durably retain a valid future transaction keyed by `(signer, nonce, tx_hash)` and return its hash without waiting for the gap to close.
- Do **not** advance the executable/pending nonce or enqueue Miden work for a gapped transaction.
- When nonce N is admitted, promote the contiguous N+1, N+2, ... sequence into the existing writer queue in nonce order.
- Preserve same-hash idempotency.
- Define a conservative conflicting-hash policy for the same `(signer, nonce)`; rejecting the conflict is sufficient for this proxy unless replacement semantics are explicitly needed.
- Bound queued work per signer and globally, with an explicit TTL/expiry outcome.
- Recover acknowledged future transactions after process restart; never return success for work that can then silently disappear.
- Keep `eth_getTransactionCount(..., "pending")` at the next nonce after the contiguous executable prefix; gapped queued transactions must not make it jump.

## Acceptance tests

- Submit N+1 before N: N+1 returns its hash promptly and its receipt remains `null`.
- Submit N: N and then N+1 are promoted/executed in order and both reach terminal receipts.
- Submit a shuffled multi-nonce burst and verify exact nonce execution order.
- Restart after acknowledging a future transaction, then fill the gap and verify recovery/promotion.
- Verify same-hash rebroadcast, conflicting same-nonce hash, queue bounds/TTL, and writer saturation without loss.
- Add a real full-path/E2E test; the current future-nonce unit test only advances the stored nonce manually.

## Non-goals

- A general public/P2P transaction pool.
- Full Geth gas-price competition and replacement policy.
- Distributed PostgreSQL multi-replica coordination; that remains separately scoped.

## References

- [Geth v1.17.4: arbitrary nonce gaps are allowed in the legacy pool](https://github.com/ethereum/go-ethereum/blob/36a7dc72e96b3f42846be925cfeb2fad18489917/core/txpool/legacypool/legacypool.go#L570-L597)
- [Geth v1.17.4: asynchronous queued-to-pending promotion](https://github.com/ethereum/go-ethereum/blob/36a7dc72e96b3f42846be925cfeb2fad18489917/core/txpool/legacypool/legacypool.go#L909-L951)
- [Geth txpool pending/queued documentation](https://geth.ethereum.org/docs/interacting-with-geth/rpc/ns-txpool)
- Prior implementation exists in divergent commit `cbc7381` (`persistent future-nonce queue`) and may be useful as reference, but should be re-evaluated against current durable-admission/recovery invariants rather than blindly restored.

mandrigin:
## Summary

Adds a bounded, persistent future-nonce holding queue so an allow-listed sender
can submit nonce-gapped transactions without the proxy holding the request open
and eventually rejecting it.

Closes #146.

## Why

If a retry at nonce `N+1` reaches the proxy before nonce `N`, rejecting or
blocking that request can prevent the sender from delivering the missing nonce.
This PR acknowledges a valid future transaction, keeps it pending, and promotes
the contiguous sequence once the gap is filled.

## RPC behavior

- Validates the signed EIP-2718 envelope before parking it.
- For `nonce > expected`, persists the envelope and immediately returns its hash
  without advancing the executable nonce or enqueueing Miden work.
- Returns `null` from `eth_getTransactionReceipt` while the transaction is parked.
- Exposes the parked transaction through `eth_getTransactionByHash` as pending.
- Keeps `eth_getTransactionCount(..., "pending")` at the contiguous executable
  frontier.
- Promotes `N+1`, `N+2`, … in nonce order after `N` is admitted.
- Treats same-hash rebroadcasts idempotently and rejects a conflicting hash at
  the same signer/nonce; gas-price replacement is intentionally unsupported.

## Persistence and recovery

- Migration `020_queued_txns.sql` stores the signed envelope and expiry block.
- Queue bounds are 256 transactions per signer and 4,096 globally.
- Parked entries use a 3,600-block TTL; Geth-style eviction makes transaction
  and receipt lookup return `null` afterwards.
- Startup and periodic recovery re-drive parked work after restart, interrupted
  promotion, or transient writer saturation.
- Promotion is serialized with live admission under the per-signer lock.

## Scope

This is a small holding/reorder pool for an allow-listed proxy, not a public P2P
txpool. It assumes one active proxy replica. Distributed PostgreSQL replica
coordination remains out of scope.

## Validation and review status

The earlier race, saturation wake-up, crash-window re-drive, and PostgreSQL bound
findings have been addressed. The PR remains draft and is **not ready to land**:

- the external E2E suite passed at `c22e888`;

- it conflicts with current `main`; the rebase must preserve both
  `020_queued_txns` and main's `021_orphan_recovery_backoff` migration;
- startup replay currently runs before HTTP bind and can wait on GER observation;
  it must retain main's bounded/background startup-recovery behavior;
- expiry currently still runs after a failed drain and can evict executable or
  recoverable work; only rows proven genuinely gap-blocked may expire;
- the dedicated future-nonce E2E still proves pending admission but does not
  assert terminal receipts, exact execution order, or a real PostgreSQL
  restart/persistence recovery path, so those regressions would not make that
  test fail;
- the normal CI job currently stops at shell lint; build and unit tests did not run.

After rebase and those focused fixes, re-review the interaction with main's #157
orphan-recovery path against the final landing diff.


mandrigin (commented):
The persistent queue is the right direction, but this is not production-ready yet. Blocking findings:

1. **The acknowledged first transaction can lose its nonce slot.** A future transaction is parked without reserving `(signer, nonce)`. After the gap-filling transaction advances the nonce, the signer lock is released before `drain_queued()` reacquires it (`service_send_raw_txn.rs:1586-1595`). A concurrent different transaction at that now-current nonce can win normal admission first. The already-acknowledged parked transaction then becomes stale and remains stranded. Preserve the first-wins invariant across promotion.

2. **Writer saturation is coupled to parking and has no recovery wake-up.** The capacity check at `service_send_raw_txn.rs:1300-1303` runs before nonce classification, so a valid future transaction is rejected rather than durably parked whenever the execution writer is full. During promotion, `drain_queued()` stops on the first transient saturation/error (`1667-1669`), but nothing retries when capacity returns. Parking must be independent of execution capacity, and contiguous durable work must resume automatically without requiring a client rebroadcast or restart.

3. **The admit-before-delete crash window is not recovered.** If a promoted transaction advances the tracked nonce and the process crashes before its queued row is deleted, that stale row is now below `expected`. Startup only drains when `peek_queued_min_nonce == expected` (`1681-1686`), so it skips the signer and can strand every later contiguous row. Clean up a stale queued row only after proving its exact hash is already durable/in-flight/terminal, then continue at `expected`; add a test that models this exact window.

4. **TTL silently forgets an acknowledged hash.** `expire_queued_txns()` deletes the only record. Afterwards both transaction lookup and receipt return `null`, with no observable terminal/expired outcome. That contradicts #146's explicit-expiry and “acknowledged work must not silently disappear” contract. Preserve an explicit terminal outcome/tombstone that RPC clients can observe.

5. **PostgreSQL bounds are not serialized.** `COUNT(*)` followed by `INSERT` in a default transaction (`store/postgres.rs:1608-1661`) is not atomic against concurrent parks for different signers, even in one proxy process, so the global cap can be exceeded. Serialize the capacity check plus insert. This is separate from the out-of-scope distributed-replica work.

Coverage must prove the production contract, not only admission bookkeeping. The current E2E uses a zero-amount proofless claim and stops after observing the pending nonce. Add terminal receipts and exact execution order for the promoted sequence, plus real PostgreSQL restart/crash-window, TTL, bounds, and writer-saturation coverage from #146.

The branch also currently conflicts with `main`; resolve that after the behavioral fixes so the final review is against the actual landing diff.

mandrigin (commented):
Re-review at `c22e888`.

The previous first-wins race, writer-capacity coupling, automatic retry,
below-expected crash-window re-drive, and PostgreSQL bound findings are now
addressed. The branch is still not production-ready, however:

1. **Startup recovery can delay service readiness indefinitely.**
   `src/main.rs` awaits `resume_queued_drain()` before binding HTTP. A replay can
   enter the strict GER-observation wait through `service_send_raw_txn`, so an
   external dependency can keep the proxy unavailable. Current `main` moved
   #157 recovery into a bounded background task specifically to avoid that.
   Preserve that invariant when rebasing this queue recovery.

2. **Expiry can delete executable/recoverable work after a failed drain.**
   `drain_queued()` invokes `expire_gap_blocked_locked()` even when replay stopped
   because the writer was saturated or another transient error occurred. The
   expiry query can then remove an expired row at `nonce == expected`, or a
   below-expected pending orphan, instead of retrying it. Only expire after the
   drain has positively established `min > expected`, and cover the
   expired-current-row + saturated-writer case.

3. **The E2E stops at pending admission.**
   `scripts/e2e-future-nonce-mempool.sh` does not assert terminal receipts for
   `N` and `N+1`, exact execution order, or promotion after a real PostgreSQL /
   process restart. Those are the acceptance properties that distinguish a
   durable queue from durable bookkeeping.

Landing notes:

- The branch conflicts with current `main` across the queue and #157 recovery
  paths. Preserve both `020_queued_txns` and `021_orphan_recovery_backoff`.
- Drop `c22e888` during rebase: its faucet-ID normalization is already on main,
  with the shell-lint typo corrected.
- Normal CI currently stops at shell lint (`unparseable` in
  `scripts/lib-isolated-wallet.sh`), so build/unit stages never ran.
- The 21 CodeQL findings are hard-coded nonce constants in tests; they appear to
  be security false positives, not production-code vulnerabilities.

Please re-request review after the minimal rebase/fixes so the final queue and
#157 orphan-recovery interaction can be reviewed as the actual landing diff.

mandrigin (commented):
Follow-up at `c22e888`: the external E2E suite passed successfully, and the PR
description now records that result.

That does **not** clear the queue-specific review findings. The dedicated
future-nonce test still does not assert terminal receipts, exact `N` then `N+1`
execution order, or promotion after a real process/PostgreSQL restart; a broken
terminal/recovery path could therefore pass the current test suite. The startup
readiness and expire-after-failed-drain correctness issues also remain, and the
branch still conflicts with current `main` across #157 recovery code.

So the external run is good integration evidence, but this PR still needs the
two minimal runtime fixes, targeted regression assertions, and a clean rebase
before landing.

IvanBelyakoff (approved):

Issue #142, opened 2026-07-14
feat(postgres): support safe active-active multi-replica proxy deployments


mandrigin:
## Context

The current PostgreSQL backend provides durable state, but the proxy execution model should be treated as **single active instance**. Process-local signer locks, tx-hash inflight state, writer queues, and claim-lock coordination do not establish one execution owner across multiple replicas sharing the same database.

PR #140 exposed the gap while hardening user-vs-sponsor claim races. Multi-replica support is explicitly out of scope for that PR/version and should be delivered as a separate deployment feature.

## Scope

- Define and document the supported topology and rollout model; until this issue closes, require one active proxy writer per PostgreSQL database.
- Add a durable execution-owner lifecycle for `(signer, nonce, tx_hash)` with explicit state, attempt/owner identity, lease or fencing token, retry/release semantics, and crash recovery.
- Ensure `HeldBy(same_hash)` cannot let two replicas execute the same raw transaction while still allowing a safe retry after an owner dies before admission.
- Make cross-replica claim ownership and terminal ClaimEvent publication fenced so a stale publisher/reclaimer cannot publish or overwrite state after losing ownership.
- Couple distributed admission with nonce/receipt outcomes so one replica cannot report acceptance while another owns the terminal result.
- Add metrics and an operator runbook for ownership, lease expiry, takeover, stale-owner rejection, and recovery.

## Acceptance criteria

- Two proxy replicas sharing one PostgreSQL database race the **same hash**: exactly one Miden publication/job executes; both callers converge on the same durable terminal result.
- They race **different hashes at the same signer nonce**: exactly one hash wins before any side effect; the loser is deterministic and consumes no work.
- Kill/restart/fail over the winning replica at every admission boundary: a new owner can recover without duplicate publication, skipped nonce, permanent reservation, or ambiguous receipt.
- A stale owner cannot commit a ClaimEvent, reverted receipt, or nonce transition after its lease/fence is superseded.
- Full two-replica PostgreSQL integration tests cover synchronous and writer-worker modes with deterministic barriers.
- Documentation clearly distinguishes durable PostgreSQL storage from supported active-active execution.

## References

- PR #140
- Follow-up to the v0.15.8 hardening work

PR #157, opened 2026-07-23
fix(recovery): replay acknowledged pending transactions after restart

### LINKED ISSUE #156: fix(recovery): automatically replay acknowledged pending transactions without a Miden handoff
## Problem

A transaction can be durably admitted by the proxy, advance the signer's accepted nonce, and return a transaction hash to the caller, while the actual writer job exists only in memory. If the process or its Miden dependency is interrupted before a durable Miden handoff is recorded, restart can leave this state:

- the nonce reservation is recorded as successfully released;
- the transaction row remains `pending`;
- no Miden transaction ID is recorded;
- no submitted note-handoff record exists;
- the in-memory writer no longer contains the job.

The durable pending-nonce frontier then correctly prevents later nonces from overtaking the unresolved transaction, but nothing automatically resumes the lower transaction. Later submissions remain rejected indefinitely unless the caller still has and re-submits the exact signed envelope. An upstream transaction manager may already have finalized or evicted its local record, so caller-driven replay is not a sufficient recovery contract.

This is a single-proxy restart/recovery problem; it does not depend on active-active PostgreSQL replicas.

## Root cause verified in current code

The reported database signature maps directly to a reachable code path:

1. `durably_admit_and_advance_nonce` creates the durable pending transaction row and CAS-advances the signer nonce.
2. The job is then accepted by the process-local Tokio writer queue.
3. Successful in-memory enqueue makes admission return `Ok`, so the nonce reservation is released as successful and the RPC hash is returned.
4. A process/dependency interruption before the worker records a durable note handoff loses the volatile queue entry.
5. A clean shutdown can produce the same state: the writer exits on the shutdown signal without draining buffered jobs, while shutdown records only a dropped-job count.
6. Startup reports dropped in-flight work but does not reconstruct and replay it from the stored signed envelope.
7. The lower pending/unlinked row then blocks admission of every later nonce.

The absence of a prepared/submitted handoff is the important retry boundary: both supported write paths persist their handoff immediately before external Miden submission. A pending row with no handoff is therefore durable proxy intent that has not crossed the external submission boundary.

The durable transaction row already contains the complete signed envelope, so it can serve as the recovery source of truth. A second durable queue should not be necessary.

## Required invariant

Once `eth_sendRawTransaction` returns a hash and the proxy advances the signer nonce, the transaction must automatically reach one of these durable outcomes without requiring client replay:

1. a submitted Miden handoff that can be reconciled; or
2. a terminal proxy receipt representing a definite outcome.

The proxy must never leave an acknowledged transaction permanently `pending` and unlinked while it blocks all later nonces.

## Minimal implementation direction

- On startup, and with a bounded periodic retry for dependency outages, find durable `pending` transactions that have no submitted/prepared handoff and no live writer job.
- Reconstruct the exact job from the stored signed envelope and resume it through the existing same-hash durable-intent path.
- Do not increment the signer nonce a second time.
- Recover multiple orphaned transactions for a signer in nonce order; do not require a one-off client replay for each row.
- Preserve same-hash ownership of the already-consumed nonce; never authorize a different hash to replace an outcome-ambiguous transaction.
- Do not treat RAM enqueue or `released_success` alone as proof that the Miden handoff completed; the durable transaction/handoff state is authoritative.
- Use bounded backoff and existing idempotency/handoff fencing so recovery cannot publish duplicate Miden work.
- Keep later nonces blocked until the lower transaction is handed off or terminal, but make the lower transaction self-healing.
- On graceful shutdown, stop producers and drain buffered writer jobs within the shutdown budget; durable recovery remains mandatory for hard crashes and incomplete drains.

## Recovery classification and backoff

The nonce must remain owned by the original hash while its outcome is ambiguous. Do not roll the nonce back or permit a different hash merely because confirmation is slow.

Classify durable pending work before acting:

- **Live writer job exists:** leave it with the active worker.
- **Prepared/submitted handoff exists:** reconcile that handoff; do not publish another transaction blindly.
- **No live writer and no durable handoff:** treat it as an orphaned durable intent and re-drive the exact stored envelope.
- **Operation is already observable in Miden state:** finalize/reconcile the existing hash rather than publishing it again.

Orphan recovery must use persistent exponential backoff with jitter and a reasonable cap. Attempt count and next-attempt time must survive process restart, retries must resume automatically when Miden becomes available, and persistent failure must alert operators. Backoff controls retry pressure; it must not turn into permanent abandonment or require the upstream client to resubmit.

Delivery is **at least once with state reconciliation**, not fragile exactly-once transport. The proxy may safely attempt the same durable GER intent more than once:

- check whether the GER is already applied before each retry;
- if it is not applied, submit the stored intent again;
- if Miden rejects it as duplicate/already applied, re-read authoritative Miden state;
- once application is confirmed, finalize the original proxy hash successfully and unblock the next nonce.

The same principle applies to claims using their authoritative claimed/global-index state. Duplicate protection in Miden is the final safety boundary, while proxy-side handoff fencing and state checks avoid unnecessary duplicate work. Deterministic validation/authentication failures must not be retried as transient outages.

## Restart reconciliation order

On every proxy restart, reconcile durable transactions for each signer in nonce order before deciding to retry:

1. **Terminal proxy receipt exists:** no recovery action is required.
2. **Miden transaction ID or prepared/submitted handoff exists:** query/reconcile the recorded Miden transaction and current bridge state. If it committed, finalize the original proxy hash; if it definitively failed, record the terminal failure; if it is still pending or the node is unavailable, retain it and poll with backoff. Do not blindly submit another transaction.
3. **No durable handoff exists:** inspect authoritative application state anyway. For a GER, check whether that GER is applied; for a claim, check the claimed global index/event. If the intended effect is already present, finalize the original proxy hash. Otherwise re-drive the exact stored envelope with backoff.
4. **Proceed to later nonces only after the lower nonce is handed off or terminal.**

A missing local response is not proof that the Miden effect is missing. Restart recovery must prefer reconciliation, then use idempotent re-submission only when reconciliation cannot find the intended effect.

## Acceptance criteria

- A proxy restart automatically recovers every pending/unlinked durable transaction without an RPC rebroadcast.
- Recovery is safe when the Miden node is temporarily unavailable and completes after connectivity returns.
- Recovery does not advance the signer nonce twice.
- Recovery cannot create duplicate Miden notes, GER injections, or claims.
- A different hash at the already-owned nonce remains rejected.
- Once the lower transaction reaches a durable handoff or terminal result, the next nonce is accepted normally.
- Startup and periodic recovery are bounded so one malformed row cannot prevent service startup or recovery of unrelated signers.

## Required regression coverage

1. **Crash after durable admission and nonce CAS, before enqueue:** restart automatically resumes the exact stored envelope.
2. **Crash after successful in-memory enqueue, before worker dispatch:** restart automatically resumes it.
3. **Crash/outage after worker dispatch begins, before durable handoff:** restart reconciles or safely re-drives without duplication.
4. **Writer-capacity race:** a pending row created before a failed enqueue remains automatically recoverable.
5. **Ordering:** a later nonce is blocked during recovery and succeeds after the lower transaction becomes durable.
6. **PostgreSQL integration test:** reproduce `pending + no Miden ID + no submitted handoff + advanced nonce`, restart the service state, and prove self-healing without client activity.
7. **End-to-end fault injection:** interrupt the proxy or Miden dependency at the admission-to-handoff boundary and prove eventual completion after restart.
8. **Backoff:** repeated dependency failures respect the persisted retry schedule; recovery resumes after restart and succeeds when the dependency returns.
9. **Graceful shutdown:** buffered and mid-flight jobs are drained when possible; any job left pending/unlinked is recovered automatically on the next boot.
10. **Multiple sequential orphans:** recover all stored envelopes in signer/nonce order without advancing any nonce twice.
11. **Ambiguous GER retry:** simulate successful GER application followed by a lost response/restart; the retry receives duplicate/already-applied behavior, verifies authoritative state, finalizes the original hash successfully, and releases the following nonce.
12. **Lost local receipt:** Miden commits a GER or claim, then the proxy restarts before recording its RPC receipt; startup reconciliation discovers the committed effect and finalizes the original hash without resubmission.
13. **Recorded handoff during outage:** restart with a prepared/submitted handoff while the Miden node is unavailable; the proxy retains and polls it with backoff rather than duplicating or abandoning it.
14. **No handoff and no effect:** restart with a pending/unlinked row whose GER/claim is absent from Miden; the proxy automatically re-drives the stored intent.

## Mandatory Miden-node restart fault tests

Add deterministic end-to-end coverage that restarts the **Miden node while the proxy is inside the Miden transaction submission call**, after proxy-side durable admission. A loose restart before or after the request is not sufficient; the harness needs a synchronization point/fault barrier around submission so the ambiguous window is exercised reliably.

Run the matrix for both supported write paths:

### GER injection

1. Persist/admit the EVM transaction and begin submitting the Miden GER transaction.
2. Restart or disconnect the Miden node before the proxy receives the submission response.
3. Cover both possible outcomes:
   - the node accepted/applied the GER but the response was lost;
   - the node restarted before accepting it, so the GER remains absent.
4. Restore the node and prove the proxy automatically reconciles or retries with backoff.
5. Assert the GER is applied exactly once, the original proxy hash reaches a terminal successful receipt, no nonce is advanced twice, and the following nonce succeeds.

### Claim

1. Prepare a valid claim with its GER already available, persist/admit it, and begin submitting the Miden claim transaction.
2. Restart or disconnect the Miden node before the proxy receives the submission response.
3. Cover both accepted-but-response-lost and not-accepted outcomes.
4. Restore the node and prove automatic claim-state reconciliation or idempotent retry with backoff.
5. Assert the global index is claimed exactly once, assets/notes are not duplicated, the original proxy hash reaches the correct terminal receipt, no nonce is advanced twice, and the following nonce succeeds.

For each path, repeat the scenario with the proxy kept alive and with the proxy restarted while the Miden node is unavailable. PostgreSQL state must be preserved throughout. The tests must require no upstream/AggKit rebroadcast and no manual database repair.

## Observability and operations

- Gauge and structured diagnostics for count and oldest age of pending/unlinked transactions.
- Counters for recovery attempts, successes, deferred retries, and terminal failures.
- Alert when a pending/unlinked transaction exceeds a configurable age.
- Document read-only diagnosis and a supported recovery command/runbook; direct database mutation must not be the normal remedy.

## Non-goals

- Active-active/multi-replica PostgreSQL coordination; tracked separately in #142.
- Nonce replacement or fee-based transaction-pool behavior.

Related: #146, because a persistent future-nonce queue depends on the same automatic recovery invariant when promoted work crosses the durable-admission-to-Miden-handoff boundary.

mandrigin:
## Summary

Adds automatic recovery for proxy transactions whose hash was acknowledged and whose signer nonce advanced, but whose in-memory writer job disappeared before a durable Miden handoff was recorded.

The signed transaction envelope already stored in PostgreSQL becomes the durable recovery source. Recovery runs at startup and every 30 seconds, preserving signer/nonce order and requiring no AggKit rebroadcast or manual database repair.

Closes #156.

## Recovery flow

```mermaid
flowchart TD
    A[Durable pending proxy transaction] --> B{Live writer job?}
    B -- Yes --> C[Leave with active writer]
    B -- No --> D{GER or claim effect already applied?}
    D -- Yes --> E[Finalize original proxy hash]
    D -- No --> F{Durable Miden handoff or transaction ID?}
    F -- Yes --> G[Reconcile with persistent backoff]
    F -- No --> H[Re-enqueue exact stored envelope]
    H --> I{Writer accepted?}
    I -- Yes --> J[Continue in signer nonce order]
    I -- No --> K[Persist exponential backoff with jitter]
```

## Implementation

- Enumerates durable pending transactions in `(signer, nonce)` order.
- Reconciles applied GERs and claimed global indices before retrying.
- Re-enqueues orphaned envelopes without advancing the proxy nonce again.
- Persists retry attempts and next-attempt timestamps in PostgreSQL.
- Adds bounded startup and periodic recovery sweeps.
- Adds backlog-age, attempt, success, defer, and persistent-failure metrics.
- Keeps active-active/multi-replica coordination out of scope under #142.

## Coverage included

- Memory-backed recovery classification and ordering tests.
- PostgreSQL reproduction of `pending + advanced nonce + no Miden ID + no handoff`.
- Deterministic post-admission fault barrier.
- E2E proxy restart and automatic self-heal without client resubmission.

## Required before ready

- Deterministic Miden-node restart during GER submission, covering accepted-but-response-lost and not-accepted outcomes.
- The same Miden-node restart matrix for a real claim.
- Verification that recorded handoffs which disappear from the restarted node reach a terminal outcome or safe retry instead of backing off forever.


mandrigin:
Review-thread cleanup: resolved 23 GitHub Advanced Security hard-coded cryptographic value annotations. Every flagged location is a deterministic integer nonce or marker inside test-only code, not a secret key or production cryptographic nonce. The aggregate CodeQL check may remain red until these alerts are dismissed or the query is tuned.

scripts/e2e-recovery-scenarios.sh
```diff
@@ -0,0 +1,314 @@
...
+    # admitted (pending row + nonce) BEFORE any Miden note handoff — the exact
+    # recoverable window. Poll the DB for that pending+UNLINKED claim rather than
+    # racing the ephemeral, undistinguished "proving CLAIM note" log line (which on a
+    # busy stack is missed or matches another claim). This also naturally tolerates a
+    # queued/slow writer (the pending+unlinked window is only longer). The wrapper's
+    # own Step-3+ timeouts are NOT our oracle; we verify recovery INDEPENDENTLY below.
+    log "  polling for MY claim to be durably admitted + unlinked (recoverable window)"
+    local hash=""
+    for _ in $(seq 1 360); do
+        hash="$(pgq "SELECT t.tx_hash FROM transactions t
```
mandrigin, line 275: Blocking for deterministic recovery evidence: this still selects the newest arbitrary pending claim, not the claim created by this deposit; since and dest are not used for correlation. There is also a check-to-fault window while bridge-service stops, and the node is restarted after a fixed 25 seconds even if the external proof has not reached submission. Please bind the row to this deposit globalIndex or calldata, freeze the proxy before the final no-handoff assertion and fault, keep the node down until target-specific failure or recovery evidence, and retain an exact balance assertion for the recovered destination. The fresh follow-up deposit does not prove the interrupted deposit was credited exactly once.

src/orphan_recovery.rs
```diff
@@ -0,0 +1,1169 @@
...
+    //    classifier syncs before reading, so a stale local view never reports a false
+    //    "absent"), using the EXACT handoff note when known so a claim landed by a
+    //    DIFFERENT transaction is not mistaken for this one succeeding (reviewer #3).
+    let outcome = match &decoded {
+        DecodedWriteCall::Ger { ger_bytes } => {
+            crate::applied_state::reconcile_ger_recovery(
+                service,
+                *ger_bytes,
+                handoff_note_id,
+                tx.handoff.is_some(),
```
mandrigin: Blocking: recover_one fetches the current handoff above, but reconciliation and the Submitted/Prepared branches still use the older scan snapshot in tx.handoff. The writer and projector do not take this signer lock, so they can create a handoff or terminalize after enumeration; recovery can then persist backoff and enqueue another expensive proof for a linked or terminal transaction. Please derive has_handoff and state from the freshly fetched handoff, treat a recorded miden_tx_id as confirmed or uncertain, and make record_recovery_attempt an atomic eligibility CAS for a still-pending row with no Miden id and no handoff. A zero-row update should stop and rescan; mirror this in memory and add a stale-snapshot race regression.

mandrigin: Addressed in d9a8f68 and re-reviewed at 5151914. Recovery now uses freshly read terminal, handoff, and Miden-tx state throughout; the PostgreSQL pre-enqueue update is an atomic eligibility gate for a still-pending row with no Miden id and no handoff, and a zero-row result stops before enqueue. The memory backend mirrors the supported single-replica semantics, with regressions for linked and terminal stale snapshots. No production regression found; resolving.

scripts/e2e-recovery-scenarios.sh
mandrigin: Re-reviewed 5151914. The pause + post-pause assertion now closes the check-to-fault race, and retaining the original wrapper gives a useful exact balance-delta oracle for the interrupted destination. Two deterministic-evidence blockers remain: (1) correlation is only by destination, but this destination is reused from the persistent e2e-suite wallet, so a stale/concurrent pending claim for that wallet can be selected; bind the exact deposit globalIndex/calldata (at minimum also require created_at >= this scenario start). (2) handoff_state=prepared is written before submit_miden_transaction, so it is not proof that submission failed; keeping the node down only until PREPARED can let the original attempt succeed after restart and bypass recovery. Require target-hash failure/backoff evidence while the node remains down, and avoid grep -q in that pipefail pipeline. Keeping this thread open for those two fixes.

mandrigin: Addressed in 36f497b. The claim scenario now excludes stale rows with a DB-clock scenario boundary plus destination/calldata correlation, freezes before the final no-handoff assertion, disables node auto-restart, keeps the node down, rejects PREPARED as evidence, and waits for target-hash failure/backoff with a pipefail-safe matcher. The original wrapper remains alive for the exact destination balance delta and the same captured hash/nonce assertions. Resolving this claim-specific thread.

scripts/e2e-recovery-scenarios.sh
```diff
@@ -0,0 +1,367 @@
...
+    since="$(date +%s)"
+    hash="$(cast send --async --rpc-url "$L2_RPC" --private-key "$GER_KEY" \
+        --legacy --gas-price 1 --gas-limit 1000000 \
+        "$BRIDGE" 'insertGlobalExitRoot(bytes32)' "$ger" 2>/dev/null)"
+    [[ "$hash" == 0x* ]] || fail "controlled GER submit not admitted (hash=$hash; REJECT_UNVERIFIED_GER=true?)"
+    log "  admitted $hash (signer=$GER_ADDR nonce=$nonce_before ger=$ger); waiting for PROOF boundary"
+    wait_for_marker "$GER_PROVE_MARKER, ger: ${ger#0x}" "$since" 180 || fail "GER never reached the proof boundary (ger ${ger#0x})"
+    assert_pending_no_handoff "$hash"
+
+    case "$fault" in node) crash_node ;; proxy) crash_proxy ;; esac
```
mandrigin: Blocking only for deterministic GER E2E evidence (production code is unaffected). The claim path now closes these races correctly, but GER still uses the old three-part pattern: `wait_for_marker` has `grep -q` under global `pipefail`; the proxy remains live between `assert_pending_no_handoff` and the fault; and the node restarts after a fixed 25s. A proof longer than 25s can submit normally after restart (so recovery was never exercised), and a fast proof can create the handoff between the assertion and kill. Please mirror the minimal claim mechanics for GER: make the marker probe pipefail-safe, pause before the final no-handoff assertion, fault while frozen, and in the node case hold Miden down until target-GER/hash failure evidence before restarting. The existing same-hash, nonce, effect, and follow-up assertions can stay unchanged.

mandrigin: Addressed in 270c679 and re-reviewed. The marker probe is now pipefail-safe; GER freezes the proxy before the final exact-hash pending/no-handoff assertion; both fault types are injected while frozen; and the node case disables auto-restart, keeps Miden down, and waits for target-hash failure evidence before restart. The same-hash success, exact nonce advance, GER-effect, recovery-counter, and next-nonce assertions remain intact. Shell syntax/diff checks pass and no production code changed; resolving.

iszubok (approved):
Approving.

**Scope of this review:** CodeQL triage. All 26 open `rust/hard-coded-cryptographic-value` alerts were examined individually and confirmed as false positives, then dismissed as *false positive* with the rationale recorded on each alert. The CodeQL check has gone from **fail → pass**.

Why they are false positives — every alert carried the same message, *"This hard-coded value is used as a nonce."*:

- The flagged literal is the **`TxLegacy.nonce` field** — an Ethereum **account sequence number**, which is public and monotonic by protocol design — not a cryptographic nonce/IV. The query's sink heuristic matches on the field name `nonce` reaching `sign_hash_sync`.
- **No key material, IV, salt, or password is hard-coded.** The signing key at every flagged site is `PrivateKeySigner::random()` (verified: all 15 key constructions in `src/orphan_recovery.rs`, plus the one in `src/store/postgres_tests.rs`).
- The other hard-coded values in the same helpers (`[marker; 32]` as a GER root, `[dest_marker; 20]` as a destination address) are ABI call payloads — public bridge data, not secrets.
- **All 26 sites are test-only code:** `src/orphan_recovery.rs` lines 759–1212 sit inside `#[cfg(test)] mod tests` (opens at line 591), and `src/store/postgres_tests.rs` is gated `#[cfg(all(test, feature = "postgres"))]`.

**Not covered by this review:** the functional correctness of the recovery logic itself (3,011 lines across 25 files — recovery ordering, nonce CAS, backoff durability). That rests on @mandrigin's testing, not on this approval.

One housekeeping note: the **"Required before ready"** section in the description looks stale — the three items appear to be addressed by later commits (`f4843a5`, `270c679`, `06a95e1`). Worth updating so the description doesn't read as blocked.

PR #159, opened 2026-08-04
feat(protocol): migrate proxy and E2E stack to Miden 0.16

mandrigin:
## Summary

Migrates the proxy and its local E2E stack to the Miden 0.16 alpha protocol
family, then carries the compatibility, monitoring, and chaos-test fixes found
while exercising that stack.

This is a **fresh 0.16 deployment migration**, not a certified proxy-only
in-place upgrade from a running 0.15 chain.

## Compatibility matrix

| Component | Version |
|---|---|
| `miden-protocol`, `miden-standards`, `miden-tx`, `miden-agglayer` | `0.16.0-alpha.5` |
| `miden-client`, `miden-client-sqlite-store` | `0.16.0-alpha.1` |
| Miden node / NTX builder / remote prover | `v0.16.0-alpha.2` |
| Miden assembly/core VM family | `0.25.x` |
| Rust | `1.96+` |

Protocol alpha.5 restores the bridge network ID as runtime account state, so it
must still match the network ID assigned to this rollup.

## Protocol migration

- Adapts account, note, transaction-record, storage-map, asset-amount, and error
  handling to the 0.16 APIs.
- Creates the bridge with separate GER injector/remover roles, currently assigned
  to the same configured GER-manager account.
- Uses public B2AGG/P2ID note behavior and removes the obsolete dummy P2ID script
  registration transaction.
- Handles the callback flag as part of the faucet account ID under 0.16.
- Removes the unreadable 0.15 `.mac` genesis accounts; proxy initialization
  provisions the bridge, service, GER-manager, and faucet accounts for the new
  chain instead.

## Monitoring and E2E hardening

- Allows the canonical 0.16 bridge-admin note scripts in the unknown-wrapper
  monitor.
- Adds checked/unchecked ownership-monitor coverage metrics so monitor blindness
  can be distinguished from a healthy zero-drift result.
- Makes the L1 claim assertion query chain state instead of trusting the submit
  command's exit code.
- Updates claim provenance, note construction, account IDs, and timeout handling
  for 0.16.
- Counts and budgets AggKit watchdog recoveries in the chaos verdict.

## AggKit lost-in-transit limitation

AggKit `0.8.3-rc1` can retain a deterministic GER transaction in its ephemeral
monitoring DB even when a proxy crash prevents that transaction from ever being
durably admitted. Stock AggKit then polls the unknown hash instead of
rebroadcasting it.

The chaos harness mirrors the current operational workaround: after detecting
that signature, it force-recreates the affected AggKit container to clear the
ephemeral monitor DB and refetch the chain nonce. A green chaos run therefore
demonstrates recovery **with this mitigation**, not native AggKit self-recovery.

## Deployment / upgrade boundary

- Use a fresh 0.16 node state and freshly provisioned proxy accounts.
- Existing 0.15 serialized account fixtures are not compatible with this stack.
- This PR does not certify swapping only the proxy binary on a running 0.15
  chain. A coordinated chain upgrade requires its own migration and recovery
  procedure/test.

## Validation and review status

- Normal lint/build/unit/PostgreSQL CI: green
- CodeQL: green
- External E2E suite: passed for this 0.16 stack at `9a590f8`; the two later
  commits are focused ownership-monitor regression coverage and version/default
  cleanup (the GitHub E2E job itself was skipped)
- Aikido: **red due to a false-positive PR license delta**. The complete
  `main..35ea8c4` license comparison introduces no new GPL/LGPL/MPL package and
  no package with missing license metadata. Newly surfaced legacy
  `MIT/Apache-2.0` manifests ship both standard license texts; the only
  weak-copyleft option in this dependency area (`priority-queue 2.7.0`, usable
  under MPL-2.0) already exists unchanged on `main`. Dismiss/allowlist the
  scanner finding as not introduced by this PR.

The previous faucet-ownership blocker is addressed: expected ownership now comes
from authoritative registration provenance before account decoding, and focused
tests cover a foreign bridge-owned faucet whose decoder degrades.

The final documentation/tooling sweep is complete: README now matches the 1.96
MSRV, `run-regression.sh` derives its node version from the Makefile while
preserving explicit environment overrides, and the Makefile documentation now
describes the 0.16 pins. The PR is ready from the code-review perspective,
subject to the normal required checks.

Nonblocking hardening follow-up: the ownership monitor should benign-skip only
registered-native + observed-native. Registered-native combined with observed
AggLayer-owned/unknown is an inconsistent state and should alert, although the
fresh-deployment admin path prevents that state today.

#160 is stacked on this branch and should be reviewed/landed after #159.


mandrigin (commented):
Review at `9a590f8`: the direction is right, but this is not release-ready yet.

1. **The ownership monitor can still become silently blind.**
   `run_faucet_ownership_check()` calls `classify_faucet_account()`, which first
   tries the complete AggLayer faucet decoder and then falls back to the base
   `FungibleFaucet` decoder. If an AggLayer-owned faucet's code commitment or
   ownership-slot layout changes, the first decode can fail while the common
   fungible portion still decodes. The account is then classified as benign
   `native_faucet`, and ownership verification is skipped—the exact silent
   failure this patch is intended to detect. The clean-native and clean-owned
   tests do not exercise that counterexample.

   Determine native-vs-AggLayer-owned from authoritative bridge registration
   metadata (`is_native`) or persist that kind locally. For a registered
   non-native faucet, any AggLayer classification/owner decode failure must be
   reported as `undecodable` (alert), never downgraded to native.

2. **The supported release/E2E paths mix 0.15 and 0.16.**
   The Makefile selects node `v0.16.0-alpha.2`, while `run-all.sh` still advertises
   protocol 0.15 and clones `v0.15.0`; `docs/RUNNING-E2E.md` still requires Rust
   1.93 and node 0.15 although this branch requires Rust 1.96. Regression and
   standalone scripts retain additional 0.15 defaults. The upgrade scripts also
   still describe a proxy-only 0.15.9→branch swap against what is now a 0.16
   node/account format. Update or retire those paths and describe this as a fresh
   0.16 deployment unless a coordinated chain migration is actually built and
   tested.

3. **The Aikido policy gate is red.**
   GitHub reports one newly introduced HIGH license issue. Dependency comparison
   points most strongly to `ark-std 0.6.0` and malformed/non-SPDX license
   metadata, but the Aikido UI must confirm the package. Explicitly approve /
   allowlist that license metadata or update the dependency; do not silently
   treat the red gate as code noise.

Operational note: the chaos changes intentionally work around AggKit
`0.8.3-rc1` losing a GER transaction in transit by force-recreating AggKit and
clearing its ephemeral monitor DB. This is a valid current runbook mitigation,
but a green soak proves recovery *with the workaround*, not stock AggKit
self-recovery. The revised PR description now makes that boundary explicit.

Normal lint/build/unit/PostgreSQL CI and CodeQL are green; the GitHub E2E job was
skipped, so record the external full 0.16 gate separately before release.

mandrigin (commented):
Follow-up to my review at `9a590f8`:

**Aikido's HIGH result is a false positive as a PR license delta, not a newly
introduced restrictive license.** I checked the complete locked dependency
comparison rather than relying on the check's one-line summary:

- there is no newly introduced GPL/LGPL/MPL package identity and no dependency
  with missing license metadata;
- newly surfaced `LicenseRef-bad-mitapache-2.0` entries use the legacy
  `MIT/Apache-2.0` spelling and their immutable published archives contain both
  standard `LICENSE-MIT` and `LICENSE-APACHE` texts (`ark-std 0.6.0` is also
  unused in the resolved all-target build graph); and
- the only weak-copyleft option in this area, `priority-queue 2.7.0`
  (`LGPL-3.0-or-later OR MPL-2.0`), already exists at the exact same version on
  `main` through `miden-package-registry`; it is not introduced here and may be
  consumed under its MPL-2.0 option.

Aikido does not expose the package name in its GitHub annotation, but neither
possible scanner interpretation represents a new #159 license risk. I no longer
consider it a merge blocker. Dismiss/allowlist it as “not introduced by this PR”
and rerun the check; no dependency replacement is required for license reasons.

The external E2E suite also passed successfully at this head. That removes the
validation caveat from my earlier review.

The two remaining blockers are unchanged and are now isolated to:

1. deriving native-vs-AggLayer-owned faucet provenance authoritatively, so a
   broken AggLayer decoder cannot fall through to benign `native_faucet`; and
2. making the supported bootstrap/docs/upgrade paths consistently 0.16 instead
   of mixing node/account-format 0.15 defaults with the 0.16 stack.

The PR description has been updated with both the successful E2E result and the
license disposition.

mandrigin (commented):
Re-reviewed `03b40df` (the two commits after `9a590f8`).

The prior functional blocker is genuinely addressed: expected faucet ownership now comes from registration provenance before the degradable decoder, and the new counterexample test covers a foreign bridge-owned faucet that decodes as a plain fungible faucet. I found no new runtime regression; normal CI and CodeQL are green, and the Aikido result remains a false-positive license delta rather than a newly introduced incompatible dependency.

One small required cleanup remains before I would call the 0.16 migration current/ready:

- `README.md:51` says Rust 1.93 while `Cargo.toml` requires 1.96.
- `run-regression.sh:12-17` still advertises/exports node `v0.15.0` (the Makefile currently overrides it, but direct use is stale/misleading).
- `Makefile:185-202` still describes the old 0.15 pins immediately above the 0.16 values.

Nonblocking hardening: only registered-native + observed-native should skip. Registered-native + observed-AggLayer/unknown should alert as inconsistent, although the fresh-deployment admin path prevents this state under the current scope.

mandrigin (commented):
Re-reviewed `35ea8c4` (`03b40df..35ea8c4`). The pushed commit fully addresses the three requested cleanup items:

- README now states the actual Rust 1.96 MSRV.
- `run-regression.sh` no longer carries a stale 0.15 pin; it derives the node URL/ref from the Makefile while preserving explicit environment overrides.
- The Makefile comments and pin consistently describe protocol/node 0.16.

The delta is documentation/tooling only, `git diff --check` is clean, and the previously reviewed faucet-provenance fix is unchanged. **Ready from the code-review perspective**, subject to required checks. Aikido remains the already-audited false-positive license delta.

The registered-native/observed-mismatch alert remains a nonblocking hardening follow-up; it does not move the merge bar for the fresh-deployment scope.

sibuser (approved):

sibuser (approved):

iszubok (approved):

PR #160, opened 2026-08-05
feat(auth): use ECDSA for newly provisioned operator accounts

mandrigin:
## Summary

Switch newly provisioned single-signature operator accounts from
`Falcon512Poseidon2` to `EcdsaK256Keccak` (secp256k1 + Keccak).

This PR is stacked on #159 (`feat/protocol-0.16`) and should land after it.

## Scope

The shared operator-account construction path now provisions ECDSA authentication for:

- the proxy service and GER-manager wallets;
- standalone and foreign-deployment wallets created by `bridge-out-tool`;
- operator-owned native faucets created by `bridge-out-tool`.

Bridge accounts and AggLayer-owned wrapped faucets continue to use their existing
role/component authorization. This prepares operator authentication for a future
ECDSA/KMS signer backend; KMS integration itself is not included here.

## Security and compatibility

- Keys are generated from an OS-backed cryptographically secure RNG.
- The auth component derives its scheme and public-key commitment from the same
  key persisted in `FilesystemKeyStore`.
- ECDSA public keys and signatures are disclosed during proving; this is an
  accepted trade-off for these operator-owned infrastructure accounts.
- Existing Falcon accounts and stored Falcon keys remain supported and are not
  migrated or rekeyed. Only newly provisioned accounts use ECDSA.

## Validation

- `make lint`, workspace build, and PostgreSQL feature check: green
- Unit tests: 501 passed, 1 ignored
- CodeQL and Aikido: green
- External stacked E2E suite: passed at `9b735fb` (the GitHub E2E job itself was skipped)

## Review status

The focused regression test now directly pins the generated secret-key scheme,
approver scheme, and public-key commitment. Provisioning uses Miden's OS-random
`new_ecdsa_k256_keccak()` constructor directly and no longer adds a separate
`rand` dependency.

The final wording cleanup is complete: the orphaned `rand` comment is removed,
existing Falcon compatibility is described accurately, and the two-key test is
correctly documented as detecting accidental key reuse rather than proving RNG
quality.

Ready from the code-review perspective behind #159, subject to the normal
required checks.

PR #162, opened 2026-08-06
feat(custody): add fail-closed Web3Signer remote signing

### LINKED ISSUE #163: feat(custody): productionize Web3Signer-backed operator signing
## Goal

Support production custody of the proxy's Miden operator-account ECDSA keys through a Web3Signer-compatible service, including AWS KMS-backed keys, without generating or persisting those private keys in the proxy.

This follows the deployment model where operators pre-create keys, pin their identifiers/commitments in proxy configuration, grant the signer the required permissions, and then start the proxy.

## Acceptance criteria

- [ ] Remote custody and local development custody are mutually exclusive and fail closed; production hardening refuses local-key fallback.
- [ ] Each operator role/account is bound to an explicitly configured key identifier/commitment—never to the first key returned by signer discovery.
- [ ] Startup verifies that every configured or restored account authentication commitment is available from the signer and preserves the exact account-to-key association.
- [ ] Creating, importing, rotating, or granting permissions to KMS keys remains an operator/runbook concern; the proxy does not generate KMS keys.
- [ ] The production proxy-to-signer boundary is authenticated and encrypted, or a strictly defined same-host/private-sidecar boundary is enforced. Hardened mode must not silently expose an unauthenticated remote HTTP signing oracle.
- [ ] Signer availability and signing failures are visible through readiness/metrics without leaking key material.
- [ ] Docker/E2E uses a supported Web3Signer release containing the Netty `CVE-2025-24970` fix and uses valid current AWS KMS configuration (`kmsKeyId`).
- [ ] Automated coverage proves successful remote signing, no local fallback, wrong/missing-key failure, restored-account binding verification, signer loss/recovery, and a production-shaped AWS KMS path before release certification.
- [ ] Deployment and key-provisioning/rotation runbooks are documented.

## Out of scope

- Proxy-side KMS/HSM key generation.
- Proxy-side cloud IAM provisioning.
- Generic transaction pooling or unrelated nonce management.

mandrigin:
## Summary

Add fail-closed remote custody for the proxy's Miden operator accounts through a Web3Signer-compatible API. The proxy can now keep ECDSA private keys in a Web3Signer backend such as AWS KMS/HSM instead of persisting them locally.

This is now rebased directly onto `main`; the prerequisite protocol/ECDSA work
from #159 and #160 has landed.

Closes #163.

```mermaid
flowchart LR
    Tx["Miden transaction\ncommitment"] --> KS["ProxyKeystore"]
    KS -->|"publicKeys + sign(digest)"| W3S["Web3Signer"]
    W3S --> KMS["KMS / HSM\nsecp256k1 key"]
    KMS --> W3S
    W3S -->|"ECDSA signature only"| KS
    KS --> MC["Miden client"]
    MC --> MN["Miden node"]

    subgraph Proxy["miden-agglayer proxy"]
      Tx
      KS
      MC
    end

    subgraph Custody["remote custody boundary"]
      W3S
      KMS
    end

    style Proxy fill:#e8f1ff,stroke:#3973ac,color:#172b4d
    style Custody fill:#f2e8ff,stroke:#7552a3,color:#2d1b4e
    style MN fill:#e8f8ed,stroke:#3b8d54,color:#173b22
```

## Behavior

- Remote custody is the production/default mode: configure `--signer-url` / `AGGLAYER_SIGNER_URL`.
- Local on-disk custody requires the explicit development-only `--insecure-local-keystore` opt-in.
- Configuring both modes, or neither, fails startup; there is no per-account fallback to local keys.
- `--require-hardening` refuses local custody.
- `--require-hardening` also requires the signer endpoint to be genuine
  loopback. HTTPS alone authenticates the signer to the proxy, not the proxy to
  the signer, so it is not treated as a custody boundary.
- The signer directory is fetched at startup. Every serving boot resolves the
  configured `service` and `ger-manager` keys, reads the existing accounts'
  actual `AuthSingleSig` commitments, requires deployed/configured/signer-held
  commitments to agree, and rebuilds the public account-to-key map before serving.
- Public keys and `r || s || v` signatures are normalized and validated before use.
- Private-key generation/import/rotation in KMS remains an operator/deployment responsibility, not a proxy responsibility.

## Test coverage

- Unit coverage for custody-mode selection, key encodings, key lookup, signature parsing/validation, signer errors, and signer-aware account authentication.
- Docker Web3Signer integration coverage that exercises a real signing path and proves remote use from the signer's request counter.
- Negative control that stops the remote signer and verifies the same flow fails instead of falling back locally.
- Restart coverage that boots the proxy from persisted state, re-verifies both
  deployed role bindings through a stable verification gauge, and drives
  another remotely signed bridge operation.
- The existing E2E topology selects custody through environment configuration so stacked overlays remain composable.

## Production-readiness checklist

- [x] Rebased cleanly onto `main` at `6bffd59`, preserving the landed 0.16/ECDSA cleanup.
- [x] Replaced implicit `first()` selection with explicit `service` and
  `ger-manager` key configuration for fresh provisioning.
- [x] Enforce role isolation on resolved public-key commitments and cover
  alternate SEC1 encodings of the same physical key.
- [x] Run full key-directory/commitment-uniqueness validation before any
  fresh-init account provisioning, so invalid/missing/aliased role keys leave
  no partial deployment.
- [x] Rebuild the account-to-key map in the serving client on every startup,
  compare existing/restored accounts' real auth commitments with configured
  role keys, and invoke verification before serving.
- [x] Enforce an honest hardened custody boundary: genuine loopback only until
  caller authentication exists; arbitrary remote HTTP/HTTPS are rejected and
  there is no insecure override under `--require-hardening`.
- [x] Added remote-signing success/failure counters on the real signing path.
- [x] Document counter alerts using `increase` windows and add signer
  failure/recovery metric assertions to E2E.
- [x] Make failure assertions delta-based and add an actual proxy restart with
  preserved accounts followed by another signed operation.
- [x] Replace formatter-dependent restart log parsing with the
  `remote_signer_verified_accounts` gauge.
- [x] Remove the no-op `--insecure-signer-transport` flag and put the hardened
  transport matrix under the library unit-test command used by CI.
- [x] Add fail-closed wrong-signer negative coverage at the keystore boundary.
  A fuller deployed-auth-slot integration negative remains useful but is not a
  runtime blocker.
- [x] Keep the signer down until the recovery baseline is captured, with an EXIT
  trap that restores it if an intermediate assertion fails.
- [x] Remove stale references to the deleted transport flag/HTTPS acceptance,
  repair the binary hardening fixture, and run binary tests in `make test-unit`.
- [x] Moved the E2E to Web3Signer `25.2.0`, which contains the Netty
  `CVE-2025-24970` fix, validated the API/metric/signature contract, and
  corrected the AWS example to use Web3Signer's `kmsKeyId` configuration.
- [ ] Run the complete external E2E/upgrade suite on the final stacked head.
- [x] Ignore the previous unprefixed one-key YAML layout and generate missing
  role keys independently.
- [x] Reject multiple YAML files for one role instead of silently selecting the
  lexicographically first key.

The proxy must consume pre-created key identifiers/commitments from configuration. Creating new KMS keys or granting KMS permissions is deliberately out of scope and belongs in the deployment runbook.

The Docker E2E is designed to prove the Web3Signer protocol path with a
file-backed key; its final external run remains pending. Actual AWS KMS custody
also remains to be certified before production use.


src/init.rs
```diff
@@ -55,10 +58,36 @@ impl From<Accounts> for AccountsConfig {
...
+// caller there is no local key to store.
+//
+// Production note: provision one signer key per account for blast-radius
+// isolation. This picks the signer's first key deterministically, so a
+// single-key signer (the e2e setup) binds every account to that key — valid,
+// but not what a production deployment should do.
+pub async fn create_auth_component(
+    keystore: &ProxyKeystore,
+) -> anyhow::Result<(AuthSingleSig, Option<AuthSecretKey>)> {
+    if let Some(commitment) = keystore.remote_commitments().first().copied() {
```
mandrigin: **Blocking for production:** do not bind accounts to `remote_commitments().first()`. That order is not an operator configuration contract, and every call currently binds the service/GER accounts to the same key. Accept explicit pre-created signer key identifiers/commitments per operator role, verify each exact key is exposed by Web3Signer, and bind that key. Key creation and IAM grants should remain outside the proxy.

src/proxy_keystore.rs
```diff
@@ -0,0 +1,327 @@
...
+        &self,
+        account_id: &AccountId,
+    ) -> Result<BTreeSet<PublicKeyCommitment>, KeyStoreError> {
+        match self {
+            Self::Local(local) => local.get_account_key_commitments(account_id).await,
+            // In remote mode the signer's keys are the only ones that can sign
+            // for any account; the account's own auth component pins which
+            // commitment is actually required, and `get_signature` rejects
+            // anything the signer does not hold.
+            Self::Remote(backend) => Ok(backend.directory.commitments().collect()),
```
mandrigin: **Blocking:** returning every signer key for every account breaks the `Keystore` account-association semantics, while the reverse lookup below always returns `None`. Preserve a public account↔commitment mapping (no secret storage is needed), return only the key bound to this account, and at startup verify that every configured/restored account's auth commitment exists in the signer. Otherwise startup merely proves that the signer has *some* key, not the required keys. This can be solved together with the explicit role pinning above.

src/remote_signer.rs
```diff
@@ -0,0 +1,467 @@
...
+
+impl RemoteSignerClient {
+    /// Builds a client for `base_url` (e.g. `http://web3signer:9000`).
+    pub fn new(base_url: &str, timeout: Duration) -> anyhow::Result<Self> {
+        let http = reqwest::Client::builder()
+            .timeout(timeout)
+            .build()
+            .context("building the remote-signer HTTP client")?;
+        Ok(Self {
+            base_url: base_url.trim_end_matches('/').to_string(),
```
mandrigin, line 68: **Blocking for hardened deployments:** `--require-hardening` currently accepts an arbitrary unauthenticated `http://` signing endpoint. KMS prevents key extraction, but an attacker who can reach this API still has a signing oracle. Keep the solution KISS: define/enforce a same-host/private-sidecar boundary or authenticated TLS/mTLS/service-mesh contract, and require an explicit insecure-dev opt-in for plain remote HTTP. Also expose signer failures/availability through readiness or metrics so loss after startup is visible.

docker-compose.web3signer.yml
```diff
@@ -0,0 +1,78 @@
...
+# Usage:
+#   ./scripts/gen-web3signer-keys.sh    # generates the key + its config
+#   docker compose -f docker-compose.e2e.yml -f docker-compose.web3signer.yml \
+#     --env-file fixtures/.env up -d
+#
+# The e2e wrapper `scripts/e2e-web3signer.sh` does all of the above.
+
+services:
+  web3signer:
+    image: consensys/web3signer:24.12.0
```
mandrigin: **Blocking for a custody/security feature:** `24.12.0` predates Web3Signer's Netty 4.1.118 upgrade for CVE-2025-24970. Please test against a supported patched release (25.2.0 is the first release carrying the fix; current stable is newer) rather than certifying this old image. Official release note: https://github.com/Consensys/web3signer/releases/tag/25.2.0

scripts/gen-web3signer-keys.sh
```diff
@@ -0,0 +1,48 @@
...
+
+# Name the file after the address purely for human traceability; Web3Signer
+# identifies keys by their public key at the API layer regardless of filename.
+KEY_FILE="$KEY_DIR/${ADDR}.yaml"
+cat > "$KEY_FILE" <<EOF
+# Web3Signer raw-file key (E2E ONLY — production uses aws-kms/azure/hashicorp).
+# Swapping this block for a vault type is the entire production delta:
+#   type: aws-kms
+#   region: eu-west-1
+#   keyId: arn:aws:kms:...
```
mandrigin: The current Web3Signer AWS KMS key-file parameter is `kmsKeyId`, not `keyId`. Please make the production-shaped example copy-safe even though actual key creation and IAM setup stay in the deployment docs/out of proxy scope.

mandrigin (commented):
The remote-custody direction is sound: custody is XOR/fail-closed, there is no local fallback, and the real Web3Signer counter plus signer-loss negative control make the protocol path credible. I would keep this draft until the production custody invariants below are addressed. The current Docker test certifies a file-backed Web3Signer path, not AWS KMS itself. Tracking the acceptance boundary in #163.

mandrigin (commented):
#162's head is unchanged at `78a6a02`, but its base #160 was force-updated to `12a85c0`. The stack is now `CONFLICTING / DIRTY`; a rebase is required before the Web3Signer fixes can be re-reviewed.

Conflicts are in `src/init.rs` and `src/bin/bridge_out_tool.rs`. During the rebase, preserve the new-base versions of #159's 0.16 pin cleanup and #160's Falcon/key-reuse wording cleanup while adapting them to the signer-aware API—the current head still contains the superseded wording/orphan comment.

All five existing Web3Signer review comments remain applicable and non-outdated; none was addressed indirectly by the base update.

docker-compose.web3signer.yml
mandrigin: Addressed in `7b6a092`: the E2E now pins Web3Signer 25.2.0, the first release carrying the Netty 4.1.118 / CVE-2025-24970 fix. The pushed validation also exercised `/upcheck`, RAW public-key decoding, the signing metric, and the 65-byte signature path on that image. Resolving.

scripts/gen-web3signer-keys.sh
mandrigin: Addressed in `7b6a092`: the production-shaped AWS KMS example now uses Web3Signer's `kmsKeyId` parameter. Resolving.

mandrigin (commented):
Superseding my temporary base-moved note: re-reviewed the newly pushed `7b6a092`.

The branch is now cleanly rebased on exact #160 head `12a85c0` and mergeable. The #159/#160 cleanup is preserved; the first ten Web3Signer patches are behaviorally unchanged apart from carrying the corrected auth-test prose. The new commit only updates Web3Signer `24.12.0 → 25.2.0` and `keyId → kmsKeyId`; its stated endpoint/key-encoding/metric/signature validation matches the integration contract. No new regression found.

Resolved as addressed:

- patched Web3Signer image / CVE finding;
- correct AWS KMS parameter.

Still blocking production readiness:

1. explicit per-role/account signer-key pinning instead of `.first()`;
2. correct account↔key mapping and restored-account startup verification;
3. an enforced authenticated/same-host signer boundary for hardened mode, plus runtime signer readiness/metrics.

Verdict: good push, **2/5 findings addressed; keep draft** pending the remaining three.

src/remote_signer.rs
```diff
@@ -0,0 +1,819 @@
...
+                ));
+            }
+            if by_role.insert(parsed, identifier.to_string()).is_some() {
+                return Err(anyhow!("--signer-key {} given twice", parsed.as_str()));
+            }
+        }
+        // Distinct roles sharing one key defeats the isolation this exists for.
+        let mut seen: BTreeMap<&str, SignerRole> = BTreeMap::new();
+        for (role, id) in &by_role {
+            if let Some(other) = seen.insert(id.as_str(), *role) {
```
mandrigin: **Blocking:** uniqueness is checked on raw identifier strings, while lookup below deliberately normalizes compressed/tagged/raw SEC1 encodings to one commitment. The same physical key can therefore be supplied in two encodings (or casing variants) and bind both roles, defeating the blast-radius isolation. Resolve every configured identifier against the loaded directory first, then reject duplicate `PublicKeyCommitment`s. Add a test where compressed and raw encodings of the same key are assigned to different roles.

src/proxy_keystore.rs
```diff
@@ -0,0 +1,453 @@
...
+                        "the remote signer does not expose the configured key {identifier:?};                          provision it in the signer (or correct --signer-key) — refusing to bind                          an account to a different key"
+                    )
+                }),
+        }
+    }
+
+    /// Fails unless every bound account's commitment is still exposed by the
+    /// signer. Run at startup so a signer that lost a key (or a restore against
+    /// the wrong signer) is caught before any traffic is accepted.
+    pub fn verify_bound_accounts(&self) -> anyhow::Result<usize> {
```
mandrigin, line 196: **Blocking lifecycle hole:** this verifier has no call site, and the map it would inspect is empty in the serving process. `bind_account` runs only while phase-1 init creates accounts; that temporary `MidenClient` is then shut down and phase 3 constructs a new `RemoteBackend`. Normal restarts skip init entirely. Rebuild the service/GER bindings on every serving startup from `AccountsConfig`, read each existing account's actual auth commitment, require it equals the configured role commitment and is present in the signer, then run this check before readiness. Cover fresh init→serving, restart, restored-account mismatch, and missing-key cases. Merely inserting the configured value without comparing the deployed account would not verify restoration.

src/remote_signer.rs
```diff
@@ -0,0 +1,819 @@
...
+                .to_string()
+        });
+    let host = host.trim_matches(|c| c == '[' || c == ']');
+    // Loopback, or a single-label host (docker-compose service name / k8s
+    // sidecar) which is not routable off the private network.
+    if host == "localhost"
+        || host == "127.0.0.1"
+        || host == "::1"
+        || host.starts_with("127.")
+        || (!host.is_empty() && !host.contains('.') && host != "0.0.0.0")
```
mandrigin: **Blocking:** this hand-rolled host test does not enforce a private sidecar boundary. `http://127.attacker.example:9000` passes `starts_with("127.")`; userinfo-shaped URLs can also confuse the extraction; and a single-label Docker/Kubernetes Service is network-reachable by other workloads, not necessarily a same-pod sidecar. Parse with a real URL parser, reject userinfo, and permit plaintext automatically only for an actual loopback IP/`localhost` (a same-pod sidecar can use loopback). Also, `https://` provides server authentication only: this client configures no client identity/token, so other reachable callers still have a signing oracle. Remote hardened deployments need an actually authenticated boundary or a strict loopback/service-mesh contract.

src/main.rs
```diff
@@ -282,6 +334,20 @@ fn check_hardening_invariants(command: &Command) -> Result<(), Vec<String>> {
         return Ok(());
     }
     let mut reasons = Vec::new();
+    if let Some(signer_url) = command.signer_url.as_deref()
+        && !command.insecure_signer_transport
```
mandrigin: `--insecure-signer-transport` currently disables this invariant *while `--require-hardening` remains enabled*. That makes the hardening claim misleading and differs from the existing treatment of other insecure flags. Keep the modes honest: hardening should reject the insecure override; development deployments that need it should drop `--require-hardening`.

scripts/gen-web3signer-keys.sh
```diff
@@ -0,0 +1,79 @@
...
+            pub="$(cast wallet public-key --private-key "$priv" 2>/dev/null)"
+            [ -n "$pub" ] || { echo "[web3signer-keys] FAIL: cannot derive public key for $role" >&2; return 1; }
+            printf '%s%s=%s' "$sep" "$role" "$pub"
+            sep=","
+        done
+        printf '\n'
+    } > "$env_file" || return 1
+    echo "[web3signer-keys] wrote $env_file"
+}
+have=$(compgen -G "$KEY_DIR/*.yaml" >/dev/null && ls "$KEY_DIR"/*.yaml | wc -l || echo 0)
```
mandrigin: **E2E migration regression (non-runtime):** the previous version leaves one unprefixed `${ADDR}.yaml`. This path sees one key, generates two role files without removing/ignoring the legacy file, then `emit_env` emits all three and turns the legacy filename into an unknown role. With two arbitrary YAML files it also falsely treats the role set as complete. Check/create exactly one file per required role and ignore or explicitly migrate legacy extras; exercise the old-one-key reuse path.

docs/operations/monitoring.md
```diff
@@ -178,6 +178,27 @@ to a plain-fungible view when the AggLayer decode fails, so trusting it would le
...
+When `--signer-url` is set, every account signature leaves this host. The proxy
+never falls back to a local key, so losing the signer stalls claims and GER
+injection outright — and the fail-closed startup check only covers boot.
+
+| Metric | Meaning |
+|---|---|
+| `remote_signer_signatures_total` | Signatures the signer produced. |
+| `remote_signer_signature_failures_total` | Signatures it failed to produce (unreachable, refused, or no key for the requested commitment). |
+
+Alert on any sustained `remote_signer_signature_failures_total`, and treat a
```
mandrigin: A counter never returns to zero after recovery, so alerting on a sustained non-zero value would page forever after one transient failure. Document an `increase(...)`/`rate(...)` window (and the success-progress condition) rather than the absolute counter value. Ideally the custody E2E should stop the signer, assert the failure counter increases, restart it, and assert successful signing resumes.

mandrigin (commented):
Re-reviewed the force-pushed `493c064`. The branch is cleanly rebased onto `main`; the previous 11 Web3Signer commits are patch-identical, and the two new commits are directionally good. The image/KMS fixes remain intact and signer success/failure metrics now cover the real signing path. However, the three production findings are only partially addressed: the serving/restart lifecycle never rebuilds or verifies the account↔key map, equivalent encodings can still assign one physical key to both roles, and the transport gate does not establish an authenticated signer boundary. Keep this draft until the blockers below are fixed. The external E2E should include a proxy restart after account creation and prove both distinct role keys are exercised.

src/init.rs
mandrigin: Partially addressed in `acaa562`: `.first()` is gone and roles are explicit. Keeping this open because isolation is checked on raw identifier strings; the same physical key in two valid SEC1 encodings/casing variants still binds both roles. The new current-line review asks to enforce uniqueness after resolving identifiers to commitments.

src/proxy_keystore.rs
mandrigin: Not yet addressed across the real lifecycle. The map is populated only in the temporary phase-1 init client's keystore, which is shut down; the serving client and every normal restart construct a fresh empty map. `verify_bound_accounts()` has no call site and does not compare restored accounts' actual auth commitments. Keeping open; the new current-line review includes the required startup/restart validation and tests.

src/remote_signer.rs
mandrigin: Metrics/observability are partially addressed in `493c064`, but the hardened transport boundary is not. The classifier trusts spoofable `127.*` hostnames and all single-label service names, and `https://` has no caller authentication in the client. The insecure override also weakens `--require-hardening`. Keeping this thread open; current-line comments contain the minimal KISS posture.

src/proxy_keystore.rs
```diff
@@ -0,0 +1,464 @@
...
+                "remote signer is reachable but exposes no usable secp256k1 key — refusing to \
+                 start with a signer that cannot sign for any account"
+            );
+        }
+        tracing::info!(
+            target: crate::COMPONENT,
+            remote_keys = directory.len(),
+            "remote signer attached — ALL account signing is remote; this process holds no key"
+        );
+        Ok(Self::Remote(Arc::new(RemoteBackend {
```
mandrigin, line 112: **Blocking fresh-init safety:** `resolve_unique()` is still not called here, before account creation. On a fresh deployment, each role is resolved independently in `create_auth_component`; a missing GER key can leave an orphan service account, and one physical key supplied in two encodings can create service+GER accounts and deploy GER/bridge state before phase-3 verification finally rejects it. Correcting config cannot repair those already-created accounts. Preflight all configured identifiers and commitment uniqueness here (or immediately before `add_accounts`) before any init side effect, and test missing-key/alias configurations fail before provisioning.

src/remote_signer.rs
```diff
@@ -0,0 +1,961 @@
...
+pub fn classify_signer_transport(base_url: &str) -> SignerTransport {
+    let Ok(url) = reqwest::Url::parse(base_url.trim()) else {
+        return SignerTransport::UnauthenticatedRemote;
+    };
+    // Credentials in the URL are never a boundary and often a typo/injection
+    // vector; refuse to treat such a URL as trusted regardless of scheme.
+    if !url.username().is_empty() || url.password().is_some() {
+        return SignerTransport::UnauthenticatedRemote;
+    }
+    if url.scheme().eq_ignore_ascii_case("https") {
```
mandrigin, line 897: **Still blocking hardened remote use:** accepting `https://` proves the proxy authenticated the server; it does not authenticate the proxy to Web3Signer. `RemoteSignerClient::new` still supplies no client identity, certificate, authorization token, or equivalent, so every other caller with network reach retains the signing oracle. KISS option: under `--require-hardening`, allow only actual loopback and place Web3Signer/an authenticated relay on the same host/pod until caller authentication is implemented. Do not label arbitrary HTTPS as an authenticated custody boundary.

scripts/e2e-web3signer.sh
```diff
@@ -0,0 +1,219 @@
...
+esac
+# Custody is either/or: the proxy must NOT also be configured for on-disk keys.
+# (The proxy refuses to start with both, so reaching a healthy state already
+# implies this — asserting it anyway keeps the guarantee visible in the log.)
+PROXY_ENV="$(docker inspect "$PROXY" --format '{{json .Config.Env}}' 2>/dev/null)"
+case "$PROXY_ENV" in
+    *AGGLAYER_INSECURE_LOCAL_KEYSTORE=true*)
+        fail "the proxy has AGGLAYER_INSECURE_LOCAL_KEYSTORE=true AND a signer — custody must be either/or" ;;
+esac
+pass "proxy started in remote-only custody mode (fail-closed startup passed)"
```
mandrigin, line 94: The exact lifecycle defect fixed by this push occurred across temporary-init client → serving client and normal proxy restart, but this E2E still only first-boots the proxy and later restarts the signer. Please restart the **proxy** with persisted account/node state, require the two role bindings to verify again, and drive another signed operation. Also add a mismatch/missing-key startup negative case (unit/integration is fine).

scripts/e2e-web3signer.sh
```diff
@@ -0,0 +1,219 @@
...
+# ── 6. the failure is OBSERVABLE, and signing RECOVERS ───────────────────────
+# A negative control that only proves "it broke" leaves the operator-facing half
+# untested: the failure must show up in metrics (that is what pages someone) and
+# the system must come back once the signer returns. A custody feature that
+# fails visibly but never recovers is not usable.
+proxy_metric() {
+    curl -s "http://127.0.0.1:8546/metrics" 2>/dev/null \
+        | awk -v m="$1" '$1 == m {print $2; found=1} END {if (!found) print 0}'
+}
+FAILS_AFTER_OUTAGE="$(proxy_metric remote_signer_signature_failures_total)"
```
mandrigin, line 209: Small test race: `> 0` can pass from an earlier failure, and the success baseline is captured only after `docker start`, so a fast recovered retry can advance before the baseline and then appear stuck. Capture both failure/success counters before stopping/starting the signer and assert deltas, or explicitly drive one post-restart signing operation after taking the baseline.

src/main.rs
```diff
@@ -253,6 +253,58 @@ struct Command {
...
+    /// production deployment should use `--signer-url` so key material lives in
+    /// a KMS/HSM. Refused by `--require-hardening`.
+    #[arg(
+        long,
+        env = "AGGLAYER_INSECURE_LOCAL_KEYSTORE",
+        default_value_t = false
+    )]
+    insecure_local_keystore: bool,
+
+    /// DANGEROUS: allow `--require-hardening` to accept a plain `http://` signer
```
mandrigin: Stale CLI help: the flag no longer lets `--require-hardening` accept HTTP; hardening now correctly rejects it. Describe this as a non-hardened development-only opt-in so operators are not told the opposite of the implemented invariant.

docs/operations/monitoring.md
```diff
@@ -178,6 +178,39 @@ to a plain-fungible view when the AggLayer decode fails, so trusting it would le
...
+
+The second condition is the one that means custody is down rather than flaky.
+Treat a zero failure rate as meaningful ONLY while
+`increase(remote_signer_signatures_total[...])` is positive — otherwise "no
+failures" may just mean nothing is being signed at all.
+
+Transport: under `--require-hardening` the signer URL must be `https://` or a
+loopback/private-sidecar host. A plain `http://` endpoint on a routable host is
+refused, because an unauthenticated signing API is a signing oracle regardless of
+where the key material lives. `--insecure-signer-transport` overrides this for
```
mandrigin: This is stale too: `--insecure-signer-transport` no longer overrides the rule under `--require-hardening`; it is rejected. Please align the runbook with the new behavior.

mandrigin (commented):
Re-reviewed `2e4150d`. This is a strong corrective push: serving startup now rebuilds the account↔key map, reads each account's actual `AuthSingleSig` commitment, compares it to the configured role commitment, verifies signer possession, and runs before serving. URL spoofing/single-label trust, the hardening override, legacy-key handling, counter alert semantics, and signer recovery visibility are also materially improved. CI/security checks are green. Two production blockers remain: role-key preflight still happens after fresh-init mutations, and arbitrary HTTPS is still accepted without authenticating the caller to the signing service. Keep draft until those are fixed; the exact restart lifecycle also needs a regression test because it was the failure mode of the previous implementation.

scripts/e2e-web3signer.sh
```diff
@@ -0,0 +1,268 @@
...
+
+RESTART_LOG="$(docker logs "$PROXY" --since "$RESTART_TS" 2>&1 | sed -e 's/\x1b\[[0-9;]*m//g')"
+case "$RESTART_LOG" in
+    *"remote custody verified"*) ;;
+    *) fail "the proxy restarted WITHOUT re-verifying its account↔signer-key bindings — this is \
+the exact lifecycle hole the verification was added to close (init-only verification)" ;;
+esac
+# It must have verified BOTH roles, not just booted.
+case "$RESTART_LOG" in
+    *"accounts: 2"*) ;;
```
mandrigin: **Blocking the new E2E evidence:** proxy logging uses `tracing_subscriber::fmt().pretty().compact()`, whose default field formatter emits `accounts=2`, not `accounts: 2`. This assertion therefore false-fails even when both bindings were verified. Match the actual stable token (`accounts=2`) or expose/assert a stable metric/event, then run this dedicated E2E externally (GitHub currently skips it).

src/main.rs
```diff
@@ -253,6 +253,63 @@ struct Command {
...
+    ///
+    /// A KMS stops key EXTRACTION; it does nothing about a reachable signing API,
+    /// which is a signing ORACLE. Note this client presents no identity to the
+    /// signer, so `https` authenticates the server to us and not us to it.
+    #[arg(
+        long,
+        env = "AGGLAYER_INSECURE_SIGNER_TRANSPORT",
+        default_value_t = false
+    )]
+    insecure_signer_transport: bool,
```
mandrigin: This flag currently changes no accepted configuration. Without hardening, `check_hardening_invariants` returns before checking transport, so a non-loopback signer is already accepted without the flag; with hardening, the flag is rejected. KISS: either remove the no-op flag or enforce the advertised explicit opt-in for every non-loopback endpoint outside hardened mode.

src/main.rs
```diff
@@ -282,6 +339,39 @@ fn check_hardening_invariants(command: &Command) -> Result<(), Vec<String>> {
...
+    }
+    // Hardening requires an actually authenticated boundary. `https` gives
+    // SERVER authentication only — this client sends no certificate, token or
+    // other identity, so every other caller that can reach the signer keeps the
+    // signing oracle. Until caller authentication exists, the only honest
+    // boundary we can enforce is genuine loopback (same host/pod), so that is
+    // what hardening demands (PR #162 review).
+    if let Some(signer_url) = command.signer_url.as_deref()
+        && miden_agglayer_service::remote_signer::classify_signer_transport(signer_url)
+            != miden_agglayer_service::remote_signer::SignerTransport::PrivateSidecar
```
mandrigin: Please put this security boundary under the unit command CI actually runs. `make test-unit` uses `cargo test --workspace ... --lib`, while this function and its tests live in the binary; the existing fixture also sets `insecure_local_keystore=true`, so its old expected reason counts are stale. Cover loopback accepted, remote HTTP rejected, remote HTTPS rejected, and the insecure override rejected under hardening.

scripts/e2e-web3signer.sh
```diff
@@ -0,0 +1,268 @@
...
+log "negative control: stopping the signer — account signing must now fail"
+# Baselines captured BEFORE the outage. Taking the success baseline after
+# `docker start` races a fast recovered retry: the retry can advance the counter
+# before the baseline is read, after which recovery looks stuck.
+proxy_metric() {
+    curl -s "http://127.0.0.1:8546/metrics" 2>/dev/null \
+        | awk -v m="$1" '$1 == m {print $2; found=1} END {if (!found) print 0}'
+}
+FAILS_BEFORE_OUTAGE="$(proxy_metric remote_signer_signature_failures_total)"
+SIGS_BEFORE_RECOVERY="$(proxy_metric remote_signer_signatures_total)"
```
mandrigin: Small remaining recovery race: this success baseline is taken before `docker stop` completes. A signature in that window makes the later recovery check pass immediately even if signing never resumes. Capture the success baseline after the outage/failure is observed and immediately before restoring the signer, then require a post-start delta.

mandrigin (commented):
Re-review of `ce313ca`: the two production-code blockers are fixed. Fresh init now validates every role and commitment-level uniqueness before `add_accounts`, and hardened custody now honestly requires a genuine loopback boundary instead of treating server-authenticated HTTPS as caller authentication. Serving-startup verification still checks the deployed auth slots.

The remaining work is small and concentrated in configuration/test assurance: one definite false-failing restart assertion, a currently no-op insecure-transport flag, hardening tests that are outside the CI unit target, and a narrow recovery-baseline race. The existing lifecycle thread remains open for the missing restored-account mismatch / true preflight-negative test; the exact-one-YAML harness thread also remains open. No new production-path architecture blocker found.

src/remote_signer.rs
mandrigin: Addressed in ce313ca: hardened mode now accepts only a genuine loopback signer or relay, rejects every non-loopback HTTP/HTTPS endpoint and the insecure override, and signer failures are exposed through counters. Resolving the original architecture finding; smaller flag/test semantics are tracked on current lines.

src/remote_signer.rs
mandrigin: Addressed in ce313ca: URL parsing remains strict and hardening now requires the parsed endpoint to classify as genuine loopback. Arbitrary HTTPS and single-label/network-reachable hosts are rejected. Resolving.

src/proxy_keystore.rs
mandrigin: Addressed in ce313ca: init_internal resolves the complete role directory and enforces commitment-level uniqueness immediately before add_accounts, so invalid, missing, or aliased keys fail before any account/deployment mutation. Resolving; focused negative coverage remains tracked in the lifecycle test thread.

src/remote_signer.rs
mandrigin: Addressed in ce313ca with the KISS boundary proposed here: hardened mode accepts only genuine loopback until caller authentication exists; arbitrary HTTPS is no longer treated as an authenticated custody boundary. Resolving.

scripts/e2e-web3signer.sh
mandrigin: The original race is addressed in ce313ca: failure checks are deltas and both counters are captured before the signer is restored. Resolving this thread; a narrower success-baseline window before docker stop is now tracked on the current line.

src/main.rs
mandrigin: Addressed in ce313ca: the CLI help now says this is development-only and explicitly states that hardening rejects the flag. Resolving.

docs/operations/monitoring.md
mandrigin: Addressed in ce313ca: the runbook now documents loopback-only hardened custody, rejection of the insecure flag, and why HTTPS without caller identity is insufficient. Resolving.

scripts/e2e-web3signer.sh
mandrigin: Partially addressed in ce313ca: the E2E now restarts the proxy on persisted state, requires both bindings to re-verify, and drives a post-restart deposit with a signature-counter delta. Keeping this thread open because the added missing-role test does not exercise init preflight or prove zero provisioning, and there is still no restored-account commitment-mismatch startup-negative case. The current-line review also catches a false-failing restart log matcher.

mandrigin (commented):
Re-review of `f48b728`: most prior assurance findings are genuinely fixed. The restart assertion now uses a stable gauge, the no-op transport flag is removed, the hardened boundary has library/CI coverage, and per-role YAML selection is exact. I found no new production-path design flaw.

One certification blocker remains: the recovery test still restores the signer before taking its claimed pre-restore baseline because an earlier `docker start` was left in place. The lifecycle negative test also remains backend-only rather than exercising the production startup verifier, and a few docs/tests are stale after removing the flag. Keep draft until these small items and the external Web3Signer E2E are complete.

scripts/e2e-web3signer.sh
```diff
@@ -0,0 +1,269 @@
...
+            SAW_FAILURE=1; break ;;
+    esac
+    sleep 2
+done
+
+# Stop driving traffic and restore the signer before asserting, so a failure
+# here never leaves the stack wedged for whatever runs next.
+kill "$NEG_PID" 2>/dev/null || true
+wait "$NEG_PID" 2>/dev/null || true
+docker start "$SIGNER" >/dev/null 2>&1 || true
```
mandrigin: **Still blocking the recovery evidence:** this starts the signer before the new baseline at line 215; the second `docker start` at line 216 is then a no-op. A pending retry can sign during lines 189–215, recreating the exact false-pass/timeout race this push intends to fix. Remove this early start and restore the signer only after capturing `SIGS_BEFORE_RECOVERY`.

docs/operations/monitoring.md
```diff
@@ -178,6 +178,46 @@ to a plain-fungible view when the AggLayer decode fails, so trusting it would le
...
+  and increase(remote_signer_signatures_total[15m]) == 0
+```
+
+The second condition is the one that means custody is down rather than flaky.
+Treat a zero failure rate as meaningful ONLY while
+`increase(remote_signer_signatures_total[...])` is positive — otherwise "no
+failures" may just mean nothing is being signed at all.
+
+Transport: under `--require-hardening` the signer URL must be a **loopback**
+address, and `--insecure-signer-transport` is **rejected** (it does not override
```
mandrigin: The flag was removed in this push, so this operator guidance is now stale. Delete the `--insecure-signer-transport` discussion and state simply that hardened custody requires a genuine loopback endpoint.

src/remote_signer.rs
```diff
@@ -0,0 +1,1045 @@
...
+/// reach the signing API from using it as a signing ORACLE — the proxy's whole
+/// authority is "can talk to the signer". `--require-hardening` previously
+/// accepted an arbitrary unauthenticated `http://` endpoint, so a hardened
+/// deployment could point at a signer across the network with no authentication
+/// and no transport integrity, and nothing would object.
+///
+/// Rather than invent an auth scheme, this enforces the KISS boundary the review
+/// asked for: either the signer is reachable only from this host / a
+/// private-sidecar address (no network exposure to authenticate), or the
+/// transport is `https` (TLS/mTLS/service-mesh terminated). Anything else is a
```
mandrigin: This rustdoc now contradicts the enforced policy: hardening rejects HTTPS and the insecure opt-in no longer exists. Please describe loopback as the only currently accepted hardened boundary; also rename/update the nearby `https_is_accepted` test comment so it says HTTPS is only a transport classification, not accepted by hardening.

scripts/gen-web3signer-keys.sh
mandrigin: Addressed in f48b728: emit_env now requires exactly one role-prefixed YAML for each required role, names every conflicting file, and refuses to select by filename order. Legacy unprefixed files remain ignored for binding. Resolving.

scripts/e2e-web3signer.sh
mandrigin: Addressed in f48b728: restart verification now sets remote_signer_verified_accounts and the E2E asserts that stable gauge instead of tracing formatter output. Resolving.

src/main.rs
mandrigin: Addressed in f48b728 with the KISS option: the no-op flag and environment setting were removed entirely. Resolving.

src/main.rs
mandrigin: Partially addressed in f48b728: the transport boundary and focused loopback/remote HTTP/remote HTTPS cases now live in the library and run under make test-unit. Keeping this open because the existing binary hardening fixture still sets insecure_local_keystore=true while several tests expect no such rejection; a full binary test run would retain stale expected reason counts.

scripts/e2e-web3signer.sh
mandrigin: Not yet addressed in f48b728: line 189 still starts the signer before the new baseline at line 215, and line 216 starts it a second time. A retry can therefore sign before the baseline exactly as before. Keeping open; the current-line blocker points to the leftover early start.

scripts/e2e-web3signer.sh
mandrigin: f48b728 improves this with a backend wrong-signer unit test, but it manually bind_account values and calls verify_bound_accounts directly. It still does not exercise verify_remote_bindings reading AuthSingleSig from a deployed/restored account, the deployed-versus-configured rejection branch, serving startup, or preflight-before-provisioning. Keeping this lifecycle coverage thread open.

mandrigin (commented):
Re-reviewed `49f84c4`. The remaining recovery blocker is fixed correctly: the signer stays down until after the success baseline, an EXIT trap restores it on premature failure, and the real restore happens exactly once. The hidden binary hardening tests are now part of `make test-unit`, their fixture is corrected, and the stale transport guidance is aligned. GitHub CI/security checks are green; no production-code blocker remains. The backend wrong-signer test is sufficient for this PR, while a full deployed-auth-slot negative integration test would be additional nonblocking assurance. Ready for review; the dedicated external Web3Signer E2E remains the final merge validation.

scripts/e2e-web3signer.sh
mandrigin: Accepted at 49f84c4. The real proxy-restart path is exercised with persisted state, a stable two-account verification gauge, and a post-restart signed bridge operation; the backend wrong-signer negative also proves fail-closed signer possession. A full deployed-auth-slot mismatch integration negative would be stronger assurance, but it is nonblocking for this PR. Resolving.

src/main.rs
mandrigin: Fully addressed in 49f84c4: the fixture now models remote loopback custody, the stale reason counts are repaired, and make test-unit runs both --lib and --bins. GitHub check passed with the binary tests included. Resolving.

scripts/e2e-web3signer.sh
mandrigin: Addressed in 49f84c4: the signer remains down through failure observation and baseline capture, then is restored exactly once. An EXIT trap provides cleanup only on premature failure. Resolving.

scripts/e2e-web3signer.sh
mandrigin: Addressed in 49f84c4: the leftover early docker start is removed; SIGS_BEFORE_RECOVERY is captured while the signer is still down, followed by the sole functional restart. Resolving.

docs/operations/monitoring.md
mandrigin: Addressed in 49f84c4: the deleted flag is no longer mentioned and the runbook states loopback-only hardened custody with no override. Resolving.

src/remote_signer.rs
mandrigin: Addressed in 49f84c4: the rustdoc now distinguishes transport classification from hardening acceptance, and the HTTPS test was renamed and explicitly asserts that hardening rejects it. Resolving.

iszubok (approved):

Issue #154, opened 2026-07-22
feat(rpc): add permissionless native-faucet registration


mandrigin:
## Problem

An external issuer can deploy a native fungible faucet on Miden, but making it bridgeable currently requires an operator to call the authenticated `admin_registerNativeFaucet` endpoint. This creates an unnecessary manual allowlisting step and prevents self-service Miden-originated assets.

#149 / PR #150 now provide the important prerequisite: the proxy imports the deployed account, requires the operator-owned native faucet kind, and derives/validates authoritative name, symbol, and decimals from public Miden state before changing bridge or database state.

## Goal

Add a permissionless JSON-RPC operation—proposed name `miden_registerNativeFaucet`—through which anyone can request registration of a valid, already-deployed Miden-native faucet.

The bridge contract/account must remain admin-controlled. “Permissionless” means the public RPC validates a deterministic request and then uses the proxy service account to submit the existing admin `ConfigAggBridgeNote`; it does **not** make arbitrary bridge configuration transactions permissionless.

## Proposed minimal design

- Expose `miden_registerNativeFaucet` without `ADMIN_API_KEY` authentication.
- Prefer an input containing only `faucet_id`.
- Import/read the faucet from the Miden node and fail closed unless it is:
  - committed and publicly readable;
  - a supported standard fungible faucet;
  - operator-owned/native, not an AggLayer-owned wrapped mint/burn faucet.
- Derive `name`, `symbol`, and `decimals` exclusively from the faucet account. Do not trust caller-supplied metadata.
- Set `origin_network` from the proxy's configured Miden network, with `scale = 0` and `is_native = true`.
- Reuse the validated registration implementation behind `admin_registerNativeFaucet`; do not maintain two registration pipelines.
- Preserve the authenticated admin method as an operator/recovery path.

### Canonical origin-token identity

The current admin API accepts a caller-selected 20-byte `origin_token_address`. A public caller must not be able to squat or overwrite another token's canonical AggLayer identity.

Before exposing the method, define and document one safe rule:

1. preferably derive the 20-byte origin identity deterministically from the Miden faucet ID using the protocol-approved mapping; or
2. if an externally selected address is required for compatibility, require verifiable authorization for that binding and atomically reject all faucet/origin conflicts.

Do not expose a permissionless API that accepts an unauthenticated arbitrary `origin_token_address`.

## Correctness and abuse resistance

- Registration is idempotent: the same faucet returns the same route/result.
[...]

PR #164, opened 2026-08-10
feat(0.16): permissionless native faucets, deterministic recovery, and chaos hardening

mandrigin:
Addresses [#154](https://github.com/gateway-fm/miden-agglayer/issues/154) and builds on [#149](https://github.com/gateway-fm/miden-agglayer/issues/149) / [#150](https://github.com/gateway-fm/miden-agglayer/pull/150). This branch also collects recovery and chaos-harness hardening discovered during extended 0.16 soak, crash, upgrade, and destructive-restore testing.

## Why

This work covers three related production-readiness areas:

1. let users register a valid Miden-native faucet without exposing bridge-admin authority;
2. reconstruct deterministic synthetic bridge history after losing both PostgreSQL and the miden-client store; and
3. make chaos/soak tests exercise and diagnose the recovery paths observed during long-running release testing.

## 1. Permissionless Miden-native faucet registration

Adds public `miden_registerNativeFaucet`. The caller supplies only an already-deployed `faucet_id`; arbitrary bridge configuration remains admin-controlled.

- Origin identity uses the reversible `EthEmbeddedAccountId` mapping between a Miden `AccountId` and the 20-byte AggLayer address.
- Name, symbol, decimals, and faucet kind are read from the deployed Miden account. Only an operator-owned `NativeFungible` faucet is accepted; wrapped/mint-burn faucets are rejected.
- Public and admin registration share an in-process single-flight guard. Public contention gets a retryable busy response instead of an unbounded queue.
- Registration checks the bridge's authoritative `faucet_metadata_map`. Exact bindings are idempotent, missing local rows can be reconciled, and conflicting bindings are rejected.
- An origin bound on-chain to an unrecorded different faucet is treated as the registry security tripwire and fails without mutation; adoption requires explicit operator review or verified restore.
- A new local route is written only after the configuration-note creator transaction commits and the expected bridge binding is observed.
- The proxy service account still signs and submits the bridge-admin note; the public caller never receives bridge-admin authority.

```mermaid
flowchart TD
    P[Public caller: faucet_id] --> L[Shared registration gate]
    A[Authenticated admin] --> L
    L --> M[Read authoritative faucet metadata]
    M --> K{NativeFungible?}
    K -->|No| X[Reject without mutation]
    K -->|Yes| B[Read authoritative bridge binding]
    B -->|Exact| R[Reconcile or return route]
    B -->|Conflict| X
    B -->|Unbound| C[Service account submits ConfigAggBridgeNote]
    C --> O{Expected binding observed?}
    O -->|No| F[Fail without success row]
    O -->|Yes| DB[Persist native route, scale 0]
```

The dedicated E2E covers unauthenticated registration, authoritative metadata, caller-input isolation, idempotency, conflict refusal, wrapped-faucet refusal, invalid-input no-mutation, concurrent callers, restart durability, local-row-loss reconciliation, and a Miden → L2B → Miden round trip using the derived origin.

## 2. Deterministic full-store recovery

Losing both stores removes local note metadata, durable joins, and synthetic projection state. Recovery now reconstructs more of that state from authoritative Miden history and reduces live/restore ordering drift.

### Shared ordering and per-block projection

- `ProjectionOrder` is the single chain-derivable comparator used by live catch-up and restore.
- `BlockProjection::project_block` is the shared per-block ORDER+EMIT unit.
- B2AGG siblings use bridge transaction/input order. Local-only transaction order is deliberately not used for GER ordering because it cannot be reconstructed after total store loss.
- Recovery replays all event families chronologically instead of using a kind-major pass.
- Same-block Claim events use content/global-index ordering for a stable served `logIndex`.
- Recovered B2AGG records retain the authoritative bridge consumer, and consumed-feed copies are excluded from replay.

```mermaid
flowchart LR
    N[Miden note visibility] --> J[Authoritative source joins]
    T[Bridge transaction inputs] --> J
    C[Client consumed/output records] --> J
    J --> O[ProjectionOrder]
    O --> P[Shared BlockProjection]
    P --> S[(Synthetic logs, receipts and GER state)]
    S --> K[Per-block seals and cursors]
```

### Restore orchestration

- Restore captures a Miden tip, re-imports configured accounts, and drives note visibility from genesis before replay.
- Node metadata is retained only for provenance-validated GER candidates rather than all public traffic.
- Existing sender/target provenance gates remain in force.
- Erased same-block B2AGG, Claim, and public GER inputs use authoritative bridge transaction data where recoverable; a private/bodyless or unreturned erased input aborts recovery instead of being omitted.
- The destructive gate fingerprints normalized Bridge/Claim logs, GER content and membership, and the order-sensitive GER hash chain; it also requires fresh post-restore traffic.
- Restore verifies that every reserved LET leaf emitted its `BridgeEvent` before finalizing the synthetic tip or recovery cursors.
- A bounded post-rebuild nonce bootstrap observes requests for 30 seconds, accepts the lowest observed nonce, and is available only during the 30-minute recovery window. Proper bridge reset/reconciliation remains the longer-term boundary.

The source-discovery and cursor drivers are not yet one implementation. Converging restore onto the continuously exercised normal projector, with bounded/windowed catch-up, is tracked in [#167](https://github.com/gateway-fm/miden-agglayer/issues/167).

### Recoverability boundary

Served per-block `logIndex` is now canonical and restore-stable. BridgeEvent transaction identity is chain-derived and stable. After total PostgreSQL loss, the original proxy-local transaction hashes for Claim/GER events and the internal global emission counter are not available from Miden history; recovery converges those fields to their deterministic forms instead of claiming unavailable byte identity.

The new ordering rules are intended for the fresh 0.16 deployment. Reusing a pre-change synthetic store requires an explicit reset/reindex or versioned migration; chain history alone cannot reproduce every old local ordering choice.

## 3. Chaos and recovery harness

- Private/tag-0 and foreign-deployment malformed-note workers run concurrently with independent must-fire deadlines.
- Attempts and results are persisted outside worker subshells; zero executions are a no-run, not a containment pass.
- Proxy restart injection no longer blindly recreates all AggKit state.
- `scripts/aggkit-preserve-heal.sh` targets the observed lost-in-transit AggKit send wedge, preserves and verifies all non-poison state, and reports success only after the exact resent transaction is durably admitted by the proxy.
- Recovery scenarios cover node/proxy interruption around GER injection and Claim submission; destructive full-store, upgrade, and chaos-soak tiers remain explicit release gates.
- A failed destructive one-shot leaves the proxy stopped on the partial store. Chaos PASS additionally requires no failed/exhausted heal, restored proxy health, and a fresh post-chaos L1 → Miden plus Miden → L1 operation.
- The event-completeness gate now matches reclaimed and deliberately deferred B2AGG notes by the same details-commitment-derived transaction identity served by the proxy. Unresolvable reclaimed identities fail certification explicitly, and same-block substitution regressions run under `make test-scripts`.

## Validation status

- At head `812728a`, GitHub `check`, formatting/clippy, Aikido, and the language CodeQL analyzers are green. The exact reclaimed/deferred same-block substitution and missing-identity regressions run under `make test-scripts`. Two CodeQL alerts on literal Ethereum account sequence nonces in tests were dismissed as test-only false positives.
- GitHub `e2e-l2l2` is intentionally skipped; the full stack runs on the external test host.
- An earlier external destructive run restored 43/43 historical GERs and preserved its rolling hash-chain value, with Bridge 17→17 and Claim 29→29. That run predates the current head and is supporting evidence, not the final release gate.
- Before 0.16 release, the current head still needs a clean deterministic recovery-scenarios run, full-store-loss run, upgrade-recovery run, and chaos soak.

## Operational boundaries

- `full-db-loss` is intentionally destructive and remains a dedicated tier rather than part of the default suite.
- Public registration is coordinated within one proxy process. Active-active/shared-PostgreSQL coordination remains out of scope and is tracked in [#142](https://github.com/gateway-fm/miden-agglayer/issues/142).
- This does not auto-register a faucet merely because a bridge-out was observed, remove the service account's bridge-admin authority, or expose arbitrary public bridge configuration.


src/restore.rs
```diff
@@ -630,6 +653,10 @@ async fn scan_bridge_out_bodies(
                     // mint-proof provenance path needs post `--reset-miden-store`.
                     let metadata = *note.metadata();
                     let details: NoteDetails = note.into();
+                    // #88: retain node-authoritative metadata for EVERY public
+                    // note (not just B2AGG/CLAIM) — the GER replay's MA#28
+                    // check needs it when the client store was wiped too.
+                    public_note_metadata.insert(details.commitment().as_bytes(), metadata);
```
mandrigin: **Blocker — this is not an exact note identity.** `NoteDetailsCommitment` explicitly excludes metadata; `NoteId` commits to both details and metadata, and this file already documents that two distinct NoteIds may share identical details. A second public note with the same GER details but different sender metadata can overwrite this entry, causing restore to join the consumed GER to the wrong provenance and either lose a legitimate GER or accept the wrong metadata. Join through the exact consumed NoteId/nullifier/transaction input, or retain candidates and fail closed on ambiguity—never last-write-wins by details alone. Also retain only GER-shaped candidates and move rather than clone the map; keeping every public note since genesis and cloning it makes restore memory unbounded in total chain traffic.

src/service_admin.rs
```diff
@@ -971,3 +971,266 @@ mod tests {
...
+/// Deriving it from the faucet id removes the choice entirely: the identity is a
+/// pure function of the account being registered, so there is nothing to squat.
+/// Two different faucets cannot collide except by finding a keccak collision,
+/// and the same faucet always derives the same address, which is what makes
+/// registration idempotent.
+pub fn derive_native_origin_address(faucet_id: AccountId) -> [u8; 20] {
+    use alloy::primitives::keccak256;
+    let mut preimage = Vec::with_capacity(NATIVE_ORIGIN_DOMAIN.len() + 64);
+    preimage.extend_from_slice(NATIVE_ORIGIN_DOMAIN);
+    preimage.extend_from_slice(faucet_id.to_hex().as_bytes());
```
mandrigin: **Blocker — protocol identity needs to be settled before routes persist.** This introduces a new irreversible proxy-local `last20(keccak(domain || ASCII hex))` identity, while the protocol dependency already exposes the reversible `EthEmbeddedAccountId` encoding for representing a Miden AccountId as 20 bytes. Issue #154 asks for a protocol-approved mapping. Please use the established conversion, or document an explicit protocol decision and add an interoperability/round-trip test proving why the token-origin identity must differ. A determinism-only unit test cannot establish compatibility.

src/service_admin.rs
```diff
@@ -971,3 +971,266 @@ mod tests {
...
+    let origin_address = derive_native_origin_address(faucet_id);
+    let origin_network = state.network_id;
+
+    // Single-flight: without this, two concurrent requests for the same faucet
+    // can both observe "not registered", both read the account, and both write —
+    // producing two bridge registrations for one faucet.
+    let _guard = REGISTER_NATIVE_LOCK
+        .get_or_init(|| tokio::sync::Mutex::new(()))
+        .lock()
+        .await;
```
mandrigin: **Blocker — this neither bounds admission nor coordinates all writers.** `.lock().await` lets an unauthenticated caller queue an unlimited number of slow registration tasks behind one global lock. The admin path also does not share this lock, and the conflict check below trusts only local DB state: after DB loss/lag, or during a concurrent admin registration, the public path can still submit a rebinding config note. Keep cross-replica support out of scope, but within one process use bounded admission (`try_lock`/bounded semaphore), share coordination with the admin path, and preflight/reconcile the authoritative bridge binding before emitting a note.

src/service_admin.rs
```diff
@@ -971,3 +971,266 @@ mod tests {
...
+    register_native_validated(
+        &state,
+        faucet_id,
+        origin_address,
+        origin_network,
+        0u8,
+        &derived_params,
+        &authoritative,
+    )
+    .await?;
```
mandrigin, line 1716: **Blocker — false success can make DB and bridge disagree.** `register_faucet_in_bridge()` currently returns `Ok(())` when `wait_for_transaction_commit()` returns `false`; this function then persists a route and reports success. A timed-out/expired config transaction leaves a durable registry row with no bridge route, and future calls short-circuit as `already_registered`. Only persist/return registered after the exact config is observed committed; otherwise return a recoverable pending/failure result and make restart reconciliation finish it safely.

mandrigin (commented):
Blocking self-review at `12be610`. The expanded recovery/chaos work is valuable and the PR description now reflects it, but this head is not ready yet. The inline comments cover the must-fix correctness and test-safety gaps. Cross-replica PostgreSQL coordination remains intentionally out of scope in #142.

scripts/e2e-permissionless-faucet.sh
```diff
@@ -0,0 +1,148 @@
...
+RESP6=$(rpc_public "{\"jsonrpc\":\"2.0\",\"id\":7,\"method\":\"miden_registerNativeFaucet\",
+  \"params\":[{\"faucet_id\":\"0xdeadbeefdeadbeefdeadbeefdeadbe\"}]}")
+echo "$RESP6" | grep -qi "error" || fail "a bogus faucet id was accepted: $RESP6"
+ROWS_AFTER=$(pgq "SELECT count(*) FROM faucet_registry;")
+[[ "$ROWS_BEFORE" == "$ROWS_AFTER" ]] \
+  || fail "a failed registration changed the registry ($ROWS_BEFORE -> $ROWS_AFTER)"
+pass "an invalid request fails with no registry change"
+
+echo ""
+pass "#154 COMPLETE — permissionless registration works, and the origin identity is derived, idempotent, conflict-safe and not caller-controllable"
```
mandrigin: This cannot claim #154 complete yet: it proves RPC/DB behavior but never uses the registered faucet for Miden→L1/L2B and the return claim. The existing full Miden-origin flow still registers through the admin RPC. Please make the normal round-trip obtain the derived origin through `miden_registerNativeFaucet` and exercise both directions; add a concurrent-call case and a restart/restore check so the durability and single-flight claims above are executable rather than inferred.

scripts/e2e-full-db-loss-recovery.sh
```diff
@@ -0,0 +1,187 @@
...
+PROJECT_DIR="$(cd "$SCRIPT_DIR/.." && pwd)"
+ENV_FILE="$PROJECT_DIR/fixtures/.env"
+
+export MIDEN_NODE_GIT_URL="${MIDEN_NODE_GIT_URL:-$(grep -m1 '^MIDEN_NODE_GIT_URL' "$PROJECT_DIR/Makefile" | sed 's/.*= *//')}"
+export MIDEN_NODE_GIT_REF="${MIDEN_NODE_GIT_REF:-$(grep -m1 '^MIDEN_NODE_GIT_REF' "$PROJECT_DIR/Makefile" | sed 's/.*= *//')}"
+export WEB3SIGNER_UID="${WEB3SIGNER_UID:-$(id -u)}" WEB3SIGNER_GID="${WEB3SIGNER_GID:-$(id -g)}"
+
+# Auto-detect the live compose project from the running proxy (chaos-harness
+# pattern) so this runs against wt-* soak stacks and plain stacks alike.
+PROXY_CONTAINER="${PROXY_CONTAINER:-$(docker ps --format '{{.Names}}' | grep -E -- '-miden-agglayer-1$' | head -1)}"
```
mandrigin: **Test-safety blocker.** This script irreversibly drops a database, but silently selects the first matching proxy when multiple compose/soak stacks are running. Require an explicit target (`PROXY_CONTAINER`/project) or refuse unless exactly one candidate exists, then verify that the selected Postgres container is the proxy's configured store before `DROP DATABASE`. A destructive test must never guess its victim.

scripts/e2e-full-db-loss-recovery.sh
```diff
@@ -0,0 +1,187 @@
...
+pgq() { docker exec "$PG_CONTAINER" psql -U agglayer -d agglayer_store -tAc "$1"; }
+
+fingerprint() {  # -> "uhc inj bridge claim hcv"
+    local uhc inj bridge claim hcv
+    uhc=$(pgq "SELECT count(*) FROM synthetic_logs WHERE topics[1] LIKE '0x65d3bf36%'")
+    inj=$(pgq "SELECT count(*) FROM ger_entries WHERE is_injected=true")
+    bridge=$(pgq "SELECT count(*) FROM synthetic_logs WHERE topics[1] LIKE '0x50178120%'")
+    claim=$(pgq "SELECT count(*) FROM synthetic_logs WHERE topics[1] LIKE '0x1df3f2a9%'")
+    hcv=$(pgq "SELECT encode(hash_chain_value,'hex') FROM service_state WHERE id=1")
+    echo "$uhc $inj $bridge $claim $hcv"
```
mandrigin: The test/body call this byte-identical history, but Bridge/Claim and injected GERs are represented only by counts. Losing one row and reconstructing a different row passes with the same count. Hash deterministic ordered identities/content for the relevant row sets (and the actual injected GER set), then compare those digests; keep the rolling GER hash-chain check as the order-sensitive assertion.

scripts/chaos-seeder.sh
```diff
@@ -98,24 +98,18 @@ fault_kill_prover() {
...
+        # observed live: recreating aggkit-l2b on an aged stack cold-resyncs
+        # into anvil's 256-state retention wall (#87) and PERMANENTLY halts
+        # bridgesync ("failed to extract bridge event data" hard-loop, L2<->L2
+        # dead). The lost-in-transit wedge this paired against is now healed by
+        # the standing PRESERVE-HEAL watchdog (soak harness): it detects the
+        # actual wedge signature (same tx hash repeating, unknown to the proxy)
+        # and recreates with every DB EXCEPT ethtxmanager-aggoracle.sqlite
+        # restored — monitor DB cleared, sync cursors + aggsender lineage kept.
+        # The seeder must not race or duplicate that; it only injects faults.
+        log "NOTE  restart-proxy pairing: lost-tx wedge (if any) is left to the preserve-heal watchdog"
```
mandrigin: Removing the destructive blanket AggKit reset makes sense, but the replacement named here is not present or invoked anywhere in this repository. As written, the repo-local chaos run can inject the known lost-send wedge and silently delegate recovery to an external harness that may not exist. Either wire the targeted preserve-heal into the versioned harness, or make wedge detection fail/report the run explicitly and document the required external component.

src/service_admin.rs
mandrigin: Partially addressed in `3b2875b`: `committed == false` now fails before the registry write. However, that is the service transaction which CREATES `ConfigAggBridgeNote`, not the later bridge transaction which consumes it and updates `faucet_metadata_map`. Five blind syncs do not prove the route is active. Please poll the exact authoritative bridge binding before persisting/returning success; timeout must leave no row and remain retryable. Keeping this open.

scripts/chaos-seeder.sh
mandrigin: Addressed in `4ec2ac3`: the repository now includes and invokes `aggkit-preserve-heal.sh`. I am resolving the missing-component thread; the new implementation has a separate fail-closed review item.

scripts/e2e-permissionless-faucet.sh
mandrigin: Addressed in `937a6c1`: the permissionless E2E now covers concurrent calls, restart durability, local-row-loss reconciliation, and a full Miden → L2B → Miden round trip using the public registration result. Resolving this thread.

scripts/e2e-full-db-loss-recovery.sh
mandrigin: The multiple-stack selection is fixed. One safety gap remains: failure to confirm that the proxy actually references this PostgreSQL target only emits a warning, then the script proceeds to `DROP DATABASE`. This destructive check must fail closed on a configuration mismatch or an unparseable store URL. Keeping this open.

scripts/e2e-full-db-loss-recovery.sh
mandrigin: Improved from counts to digests, but the Bridge/Claim digest currently includes only `transaction_hash`, `log_index`, and `data`. It omits block identity/number, address, topics, and removal state; indexed event regressions can therefore pass. Please fingerprint the full normalized recoverable log row. Keeping this open.

src/service_admin.rs
mandrigin: Addressed in `bc2475d`: the public path now uses the established reversible `EthEmbeddedAccountId` conversion, with round-trip and distinct-ID coverage. Resolving this thread.

src/restore.rs
mandrigin: Partially addressed in `727591e`: memory is now bounded to GER-shaped candidates and differing metadata fails closed instead of last-write-wins. The identity issue remains: an intentional public clone of a legitimate GER note's details can make the genuine entry ambiguous and omit it during full restore. This is not a random hash collision. Please join metadata to the exact consumed NoteId / transaction input. Keeping this open.

src/service_admin.rs
mandrigin: Addressed in `7d45e8b`: public admission now uses `try_lock`, the admin path shares the same lock, and both paths preflight the bridge binding before emitting a new note. Cross-replica coordination remains #142. Resolving this thread.

scripts/e2e-full-db-loss-recovery.sh
```diff
@@ -0,0 +1,268 @@
...
+#     the original interleaved arrival order, so the indices are re-numbered even
+#     when every event is faithfully reproduced.
+#
+# What MUST be identical is the part that carries meaning: each injected GER's
+# VALUE, its per-block attribution, its ORDER (asserted separately and exactly by
+# the rolling `hash_chain_value`) and its count. Those are asserted strictly.
+# Bridge/Claim keep the FULL digest (tx identity included) — theirs is
+# reconstructed from node-authoritative note bodies, so it must survive.
+uhc_content_digest() {
+    pgq "SELECT md5(coalesce(string_agg(block_number || ':' || data, '|' \
```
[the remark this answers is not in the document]
mandrigin: @AikidoSec ignore: this is a non-adversarial test equality fingerprint over database rows, not a security digest or trust boundary. A stronger hash would not materially improve this regression assertion.

scripts/chaos-garbo.sh
```diff
@@ -199,52 +205,114 @@ garbo_foreign_claim() {
...
+        # Each class is independently guaranteed. A class that fired 0 times
+        # cannot assert its containment — that is NO-RUN for that class, never a
+        # containment PASS.
+        [ "$pf" -ge 1 ] && echo "GARBO_PRIVATE_GUARANTEE=met" || echo "GARBO_PRIVATE_GUARANTEE=missed"
+        [ "$ff" -ge 1 ] && echo "GARBO_FOREIGN_GUARANTEE=met" || echo "GARBO_FOREIGN_GUARANTEE=missed"
     } > "$GARBO_SUMMARY"
-    glog "summary -> $GARBO_SUMMARY (private=$PRIVATE_FIRED foreign=$FOREIGN_FIRED)"
+    glog "summary -> $GARBO_SUMMARY (private=$pf/$pa foreign=$ff/$fa)"
 }
 trap write_summary EXIT
```
mandrigin: **Harness blocker — stopping the parent does not stop these new background workers.** `e2e-chaos-soak.sh` kills only the `chaos-garbo.sh` PID, while this EXIT trap only writes a summary. The private/foreign workers can continue for up to 900 seconds, injecting traffic during the post-chaos settle and verdict. Track both child PIDs in cleanup, forward termination, reap both, and write the summary only after they are stopped.

scripts/aggkit-preserve-heal.sh
```diff
@@ -0,0 +1,86 @@
...
+# ── the heal: preserve every DB except the poisoned monitor DB ───────────────
+B="$(mktemp -d)"; trap 'rm -rf "$B"' EXIT
+docker stop "$C" >/dev/null 2>&1
+saved=0
+for f in aggsender bridgel2sync L1InfoTreeSync reorgdetectorl1 reorgdetectorl2; do
+    for ext in sqlite sqlite-wal sqlite-shm; do
+        docker cp "$C:/tmp/$f.$ext" "$B/" >/dev/null 2>&1 && saved=$((saved+1)) || true
+    done
+done
+COMPOSE_PROJECT_NAME="$PROJECT" docker compose "${COMPOSE[@]}" --env-file "$ENV_FILE" \
```
mandrigin, line 126: **Blocker — this heal can destroy exactly the state it claims to preserve.** Every `docker cp` failure is ignored, the filenames are a hard-coded allowlist, and `saved` may be zero or `restored < saved`; the script still force-recreates the only source container, starts it, reports success, and the soak caller ignores the return code. Before recreation, require and verify the critical DB set was staged; after recreation, require every staged file (and usable ownership) was restored before start. `saved == 0` or any mismatch must fail closed, and the watchdog must count a heal only after success.

src/service_admin.rs
```diff
@@ -971,3 +1030,463 @@ mod tests {
...
+        anyhow::anyhow!(
+            "miden_registerNativeFaucet: another native-faucet registration is in progress; \
+             retry shortly (bounded admission — requests are not queued)"
+        )
+    })?;
+
+    // ── conflict checks BEFORE any state change ──────────────────────────────
+    // Idempotent for the same faucet; explicit failure for anything that would
+    // change an existing binding.
+    if let Some(existing) = state.store.get_faucet_by_id(faucet_id).await? {
```
mandrigin, line 1499: **Blocker — the authoritative preflight is bypassed by this local-row fast path.** A stale/partial row (including one created by the creator-tx false-success case) is trusted forever without checking that the bridge still has the exact binding. The admin path has the same early return at lines 400–414, and `preflight_bridge_binding` reads cached bridge state without an explicit sync. Make idempotent success chain-checked: sync/read the bridge first, then reconcile or reject the local row consistently.

scripts/e2e-permissionless-faucet.sh
```diff
@@ -0,0 +1,253 @@
...
+#   SKIP_PERMISSIONLESS_ROUNDTRIP=1 skips step 9 when the suite already runs the
+#   permissionless round-trip as its own tier.
+# ══════════════════════════════════════════════════════════════════════════════
+set -uo pipefail
+SCRIPT_DIR="$(cd "$(dirname "${BASH_SOURCE[0]}")" && pwd)"
+PROJECT_DIR="$(cd "$SCRIPT_DIR/.." && pwd)"
+FIXTURES_DIR="$PROJECT_DIR/fixtures"
+source "$SCRIPT_DIR/lib-l2l2.sh"
+
+AGG_C="l2l2-miden-agglayer-1"
```
mandrigin: **E2E blocker — this ignores the compose project/container detected by `lib-l2l2.sh`.** The suite can run under a non-`l2l2` project name, but step 8 will restart this hard-coded container or fail while the actual proxy remains untouched, so the durability assertion is not trustworthy. Use `$AGGLAYER_CONTAINER` from the shared library throughout.

scripts/e2e-full-db-loss-recovery.sh
```diff
@@ -0,0 +1,268 @@
...
+    [[ "$READY" == "True" ]] && break
+    (( SECONDS >= deadline )) && fail "post-restore deposit never ready_for_claim in 300s — pipeline dead after restore"
+    sleep 5
+done
+pass "post-restore deposit ready_for_claim"
+
+deadline=$((SECONDS + 240))
+while :; do
+    INJ2=$(pgq "SELECT count(*) FROM ger_entries WHERE is_injected=true")
+    [[ "$INJ2" -gt "$INJ1" ]] && break
```
mandrigin: **E2E blocker — this compares a count to the post-restore injected-set MD5.** `INJ1` comes from `fingerprint()`, while the numeric count is `NINJECTED1` from `counts()`. The arithmetic test is therefore invalid and the advertised fresh-GER liveness gate cannot pass reliably. Compare `INJ2 > NINJECTED1` and report those two counts.

src/service_send_raw_txn.rs
```diff
@@ -2603,6 +2640,87 @@
...
+    #[tokio::test]
+    async fn finding90_live_ledger_does_not_bootstrap_a_gap() {
+        let service = create_test_service();
+        let store = service.store.clone();
+        // No marker set — this is a normal running deployment.
+        let calldata = insertGlobalExitRootCall {
+            root: FixedBytes::from([0xBBu8; 32]),
+        }
+        .abi_encode();
+        let (input_hex, signer) = encode_legacy_tx_with_nonce(calldata, 21);
```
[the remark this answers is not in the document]
mandrigin: False positive: `21` is an Ethereum transaction sequence nonce used to test that a normal live ledger does not adopt a future gap. It is test input, not cryptographic material or a secret.

src/service_send_raw_txn.rs
```diff
@@ -2603,6 +2640,87 @@
...
+    async fn finding90_rebuilt_ledger_bootstraps_continuing_signer() {
+        let service = create_test_service();
+        let store = service.store.clone();
+        store.set_nonce_ledger_rebuilt(true).await.unwrap();
+
+        let calldata = insertGlobalExitRootCall {
+            root: FixedBytes::from([0xAAu8; 32]),
+        }
+        .abi_encode();
+        let (input_hex, signer) = encode_legacy_tx_with_nonce(calldata, 21);
```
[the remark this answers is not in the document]
mandrigin: False positive: `21` is an Ethereum transaction sequence nonce used to exercise post-restore ordering in a unit test. It is not cryptographic material, an IV, a salt, or a secret.

src/service_admin.rs
mandrigin: Addressed in `d33fe87`: after the creator transaction commits, registration now syncs and polls `faucet_metadata_map` for the exact faucet/origin/network binding. Timeout aborts before the registry write, so this false-success path is closed. Resolving.

src/restore.rs
mandrigin: Addressed in `ddce039`: the node scan now admits GER metadata only when the same MA#28 predicate proves the consensus-stamped sender and bridge target. Miden stamps the sender from the executing native account, so an external same-details clone cannot enter or poison the ambiguity map. Exact NoteId joining remains useful hardening, but the exploitable restore-griefing case is closed. Resolving.

src/service_admin.rs
mandrigin: The chain sync/check is added, but stale-row healing is still broken. In the public `Unbound` branch, control falls through and the immediately following origin lookup finds the SAME stale row and rejects it, so the advertised re-drive is unreachable. In admin, a stale row for faucet A can fall through while the caller requests B; the bridge binds B, then the guarded origin upsert affects zero rows and still reports success, leaving PostgreSQL on A. Explicitly handle/remove/repair the stale row and verify the exact persisted binding before success. Keeping this open.

src/service_send_raw_txn.rs
```diff
@@ -1374,6 +1374,43 @@ pub async fn service_send_raw_txn(service: ServiceState, input: String) -> anyho
...
+        // those apart. So restore sets a durable marker, and here — only for a
+        // signer with NO row at all, on a store flagged as rebuilt — we adopt the
+        // first observed nonce as that signer's baseline.
+        //
+        // Safety: this cannot enable replay of a LOWER nonce (seeding high makes
+        // every lower nonce stale, which R4 still rejects), and from the seed
+        // onward ordinary R4 sequencing applies. Double-EFFECTS remain prevented by
+        // the domain-level idempotency that DOES survive a restore
+        // (`is_ger_injected`, `claimed_indices` / `is_note_processed`).
+        if tx_nonce > 0 && service.store.is_nonce_ledger_rebuilt().await? {
```
mandrigin: **Blocker — post-rebuild bootstrap assumes the first arriving nonce is the lowest outstanding nonce.** HTTP/replay ordering can deliver `N+1` before still-pending `N`; this immediately seeds and accepts `N+1`, after which `N` is permanently stale. That bypasses the 30-second future-nonce ordering protection directly below. The marker is also global, set after every restore, and never cleared, so it applies indefinitely to every previously unseen signer rather than only known continuing signers from an actually empty ledger. Scope/clear recovery mode and avoid treating first arrival as proof of the baseline (or document and test a real one-in-flight-per-signer invariant). Add the `N+1`-before-`N` recovery regression; the current tests exercise only one first transaction.

scripts/e2e-full-db-loss-recovery.sh
```diff
@@ -0,0 +1,569 @@
...
+RESTORE_LOG="/tmp/full-db-loss-restore-${RUN_SUFFIX}.log"
+set +e
+docker compose "${COMPOSE[@]}" --env-file "$ENV_FILE" \
+    run --rm --no-deps miden-agglayer \
+    "${LIVE_CMD[@]}" --reset-miden-store --restore \
+    >"$RESTORE_LOG" 2>&1
+RESTORE_EXIT=$?
+set -e
+say "restore one-shot exit=$RESTORE_EXIT (log: $RESTORE_LOG)"
+if [[ "$RESTORE_EXIT" -ne 0 ]]; then
```
mandrigin, line 426: **Blocking destructive-test safety:** a failed one-shot must not restart the normal serving proxy on a partial store. Tier 1/2 share ordering/emission, but live discovery/cursors are not yet the restore pipeline (#167), and `/health` does not prove projector/reconcile catch-up. This can expose incomplete history; because `nonce_ledger_rebuilt` was not finalized, the next drill can even mislabel the reconstructed store as a fresh live baseline. Leave the proxy stopped (or run an explicit non-serving recovery-to-captured-tip gate) and fail with repair instructions.

src/restore.rs
```diff
@@ -723,13 +921,139 @@ async fn restore_bridge_replay(
...
+    ::metrics::counter!("restore_erased_claims_recovered_total").increment(claims as u64);
+    ::metrics::counter!("restore_erased_b2agg_recovered_total").increment(b2aggs as u64);
+    tracing::info!(
+        candidates = missing.len(),
+        claims,
+        b2aggs,
+        "restore: recovered erased (same-block-consumed) bridge-input note bodies \
+         from input headers"
+    );
+    if gers > 0 || opaque > 0 || unreturned > 0 {
```
mandrigin: **Blocking full-store recovery:** this detects a public same-block-created-and-consumed `UpdateGerNote`, increments `gers`, then only warns and returns `Ok`. With both stores wiped, that note is absent from the ordinary consumed feed, so its `UpdateHashChain`, `is_injected`, and rolling hash-chain contribution are silently lost while restore can still finalize successfully. Route the fetched body/metadata into the shared per-block projection stream, or at minimum abort restore whenever `gers > 0` until that is implemented. Please add the exact erased-GER recovery test.

src/restore.rs
```diff
@@ -435,71 +473,36 @@ pub async fn restore(
...
-    total_logs += ger_logs;
-    tracing::info!("Phase 3 complete: {gers} GERs, {ger_logs} logs");
+    let (bridge_outs, claims, gers) = (replayed.bridge_outs, replayed.claims, replayed.gers);
+    total_logs += bridge_outs + claims + gers;
+    tracing::info!(
+        "Phase 2 complete: {bridge_outs} bridge-outs, {claims} claims, {gers} GERs — \
+         emitted through the SHARED per-block projection unit in canonical order"
+    );
 
     let accounted = store.get_accounted_deposit_count().await?;
```
mandrigin, line 507: **Blocking recovery correctness:** restore verifies only `accounted == let_leaves`, but every replay key is reserved before projection and `project_b2agg_note` has several post-reservation `Skipped` paths (unparseable/no asset/unknown faucet/oversize/self-target). A skipped leaf therefore satisfies this count, then restore finalizes the tip/cursors with no `BridgeEvent`, creating the exact permanent `depositCount` gap the live emitted-frontier gate prevents. Run `first_unemitted_reservation()` after replay and before cursor finalization, fail the one-shot if any exists, and add a restore regression for a reserved-but-unemitted leaf.

scripts/e2e-chaos-soak.sh
```diff
@@ -124,10 +124,26 @@ WATCHDOG_HEALS_FILE=/tmp/chaos-watchdog-heals; : > "$WATCHDOG_HEALS_FILE"
...
+      # fail-closed staging guard) still consumed heal budget and still read as a
+      # successful intervention. Log the attempt separately from the outcome, and
+      # only the OUTCOME feeds the counted ledger.
+      echo "$(date +%H:%M:%S) WATCHDOG-ATTEMPT: $AK wedged on lost-in-transit tx $tx (unknown to proxy) — preserve-healing $svc"
+      if PROJECT="$PROJECT" FORCE=1 "$SCRIPT_DIR/aggkit-preserve-heal.sh" "$svc" >/dev/null 2>&1; then
+        echo "$(date +%H:%M:%S) WATCHDOG: preserve-heal of $svc SUCCEEDED (tx $tx)" \
+            | tee -a "$WATCHDOG_HEALS_FILE"
+      else
+        rc=$?
+        echo "$(date +%H:%M:%S) WATCHDOG-FAILED: preserve-heal of $svc returned rc=$rc (tx $tx) — NOT counted as a heal"
```
mandrigin: **Chaos verdict can still false-green.** A preserve-heal failure is only logged after the tx is marked `seen`, so it is never retried; heal-budget exhaustion, failure to regain proxy health, and a dead NTX builder likewise never feed `LEGIT_OK`/`CHAOS_OK`. The pre-chaos load may already have completed and the log verifier can still read old state, so this does not prove post-fault liveness. Persist watchdog failures/budget exhaustion and health failure into the final verdict, and require at least one fresh two-way post-chaos operation (or equivalent deterministic recovery scenario) before PASS.

scripts/verify-event-completeness.sh
```diff
@@ -257,22 +257,42 @@ for name, (topic, root) in TOPICS.items():
     # the proxy's refused set is DEFERRED (expected on the live path; recovery is via
     # --restore), not missing. Only unexplained absences remain in `missing`.
```
mandrigin: **Release-gate false green:** reclaimed notes are removed only after `missing > 0`. If the proxy incorrectly emits a `BridgeEvent` for a sender-reclaimed note, the initial expected count still includes that note, the incorrect log satisfies it, `missing == 0`, and this consumer check never runs—so the verifier passes the bug this change is meant to distinguish. Resolve each note's canonical consumer before building `note_blocks`: exclude only proven non-bridge consumptions from expected events, so any emitted log for them is counted as `extra`; unresolved consumers must remain fail-closed.

src/service_send_raw_txn.rs
```diff
@@ -2603,6 +2742,173 @@
...
+        // Scope the guard to the SYNCHRONOUS section touching the shared statics:
+        // holding a std Mutex across an await is a clippy error and a real deadlock
+        // hazard. The window-expiry path clears this map, so the two tests that use
+        // it must not interleave here.
+        let baseline = {
+            let _guard = NONCE_RECOVERY_TEST_GUARD
+                .lock()
+                .unwrap_or_else(|e| e.into_inner());
+            // Observed out of order: the higher nonce arrives first.
+            note_recovery_nonce(addr, 22);
```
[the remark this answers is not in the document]
mandrigin: False positive: these literals are Ethereum account sequence nonces in a unit test (`N+1` arriving before `N`), not cryptographic nonces, IVs, salts, keys, or secrets. Dismissing as used in tests.

src/service_send_raw_txn.rs
mandrigin: Accepted for this PR: recovery mode is now time-bounded, waits through the normal 30s ordering window, seeds from the lowest observed nonce rather than first arrival, and clears durably. A nonce delayed beyond that window remains an explicit heuristic boundary and belongs with the follow-up nonce/reset work, not a blocker here.

src/service_admin.rs
mandrigin: The public same-faucet stale-row re-drive is now fixed. The admin different-faucet case is still unsafe: an origin row for faucet A can fall through while the caller requests B, bind B on-chain, then the guarded upsert preserves A and the new read-back errors only after the bridge mutation. Detect/reject that stale-origin conflict before emitting a config note (or perform an explicit safe repair), and reconcile the local row on `AlreadyBound`. Keeping open.

scripts/aggkit-preserve-heal.sh
mandrigin: The staging/count and watchdog-success handling are improved, but this remains blocking. The helper still hard-codes five basenames while the pinned AggKit v0.8.3-rc1 config also has `bridgel1sync.sqlite`, `l2gersync.sqlite`, and `certificates/` state (see https://github.com/agglayer/aggkit/blob/v0.8.3-rc1/config/default.go). Those are lost on recreate. It also deletes the staging directory on a post-destructive recreate/start failure and does not actually verify ownership or process health. Please stage a manifest of all runtime state except the exact poisoned DB, retain it on every post-destructive error, restore/verify it completely, and require health before reporting success. Keeping open.

src/service_send_raw_txn.rs
```diff
@@ -2603,6 +2742,173 @@
...
+            note_recovery_nonce(addr, 21);
```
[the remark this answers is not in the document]
mandrigin: False positive: these literals are Ethereum account sequence nonces in a unit test (`N+1` arriving before `N`), not cryptographic nonces, IVs, salts, keys, or secrets. Dismissing as used in tests.

scripts/e2e-full-db-loss-recovery.sh
mandrigin: Addressed: the fresh-GER liveness comparison now uses the numeric pre-restore injected count (`NINJECTED1`) rather than the set digest. Resolving.

scripts/e2e-full-db-loss-recovery.sh
mandrigin: Addressed: the gate now fingerprints normalized recoverable log content (block identity, address, topics, data, tx index/removal state as applicable) and also compares the consumer-facing paginated `eth_getLogs` stream. Resolving.

scripts/chaos-garbo.sh
mandrigin: Addressed: TERM/INT now route through `stop_workers`, both children are killed and reaped, and the summary is written only after cleanup. Resolving.

scripts/e2e-full-db-loss-recovery.sh
mandrigin: Selection and host parsing improved, but the destructive guard is not complete yet: it accepts `localhost`/`127.0.0.1` as proof of the separate compose Postgres target, and it never parses/verifies the configured database name before unconditionally dropping `agglayer_store`. In this compose topology localhost is the proxy container itself, not `$PG_CONTAINER`. Require both the exact reachable host/service and `dbname=agglayer_store` (or URL path), otherwise fail closed. Keeping open.

scripts/e2e-permissionless-faucet.sh
mandrigin: Addressed: restart/readiness now use the container discovered by the shared L2↔L2 harness rather than a hard-coded compose project. Resolving.

scripts/e2e-full-db-loss-recovery.sh
mandrigin: Addressed in 2af3732. The destructive guard now parses and requires both the exact PostgreSQL host/service and dbname=agglayer_store; localhost, 127.0.0.1, mismatches, and unparseable values fail closed unless the operator supplies the explicit override. Resolving.

scripts/e2e-full-db-loss-recovery.sh
mandrigin: Addressed in 2af3732. A failed one-shot now leaves the proxy stopped on the partial store and prints the repair/rerun path; it no longer starts normal serving or contaminates the next fidelity baseline. Resolving.

src/restore.rs
mandrigin: Addressed in 2af3732. Public erased GER bodies retain authoritative metadata, enter the shared internal replay stream, and dispatch through project_ger_note; private/bodyless or unreturned erased inputs now abort recovery. The new regression verifies GER emission and injected state. Resolving.

src/restore.rs
mandrigin: Addressed in 2af3732. verify_emitted_frontier runs after replay/accounting and before cursor/tip finalization, with a reserved-but-unemitted regression. Resolving.

scripts/e2e-chaos-soak.sh
mandrigin: Addressed in 2af3732. Failed heals and exhausted heal budget are durable verdict inputs; proxy health is required; and PASS now requires a fresh post-chaos L1 to Miden plus Miden to L1 operation. Resolving.

scripts/aggkit-preserve-heal.sh
mandrigin: The whole-state manifest and retained staging are good improvements, but this is still blocking. docker stop is unchecked, so the script can snapshot live SQLite and then destroy its source. The pinned AggKit image runs as appuser, while both docker cp calls omit --archive; the local-to-container restore therefore creates root-owned SQLite files that may not be writable. Read-back checks only existence, not content. Finally, the health pattern accepts any RestartCount and fresh error/crash-loop logs as healthy. Require a confirmed stop, preserved and verified UID/GID plus content hashes, RestartCount 0 or no growth, and a functional progress/health probe; stop the recreated service on failed validation. Keeping open.

src/service_admin.rs
mandrigin: The original stale-origin/different-faucet mutation is fixed, but one symmetric pre-mutation conflict remains: admin never checks get_faucet_by_id(requested_faucet_id). If that faucet already exists locally under another origin, bridge preflight can be Unbound, the irreversible ConfigAggBridgeNote is emitted, and only then the faucet_id primary-key conflict makes persistence fail. Mirror the public path by-ID check before any Miden mutation. Also, stale local A versus authoritative bridge B currently returns Ok(B) while leaving PostgreSQL on A; reconcile safely or fail instead of reporting success. Keeping open.

scripts/verify-event-completeness.sh
mandrigin: Up-front consumer filtering fixes the single-note case, but the release-gate false green remains for same-block substitution. Expected notes and logs are still reduced to Counter(block): one legitimate B2AGG missing plus one wrongly emitted reclaimed B2AGG in the same block yields expected=1/logs=1, missing=0, extra=0. Deferred reclassification also re-queries all notes, including filtered reclaims. At minimum derive the forbidden BridgeEvent transaction hash from every reclaimed NoteId and fail if any such log exists, restrict deferred candidates to the filtered expected set, and add the exact same-block substitution regression. Keeping open.

scripts/e2e-full-db-loss-recovery.sh
```diff
@@ -0,0 +1,585 @@
...
+dump_rows '0x65d3bf36' "/tmp/fdl-uhc-before-${RUN_SUFFIX}.txt"
+if getlogs_dump "/tmp/fdl-getlogs-before-${RUN_SUFFIX}.tsv"; then
+    say "before: eth_getLogs captured ($(wc -l < "/tmp/fdl-getlogs-before-${RUN_SUFFIX}.tsv") logs) from $PROXY_RPC"
+else
+    fail "eth_getLogs baseline capture FAILED against $PROXY_RPC — refusing to run a drill \
+whose consumer-level comparison cannot be made (set PROXY_RPC)"
+fi
+say "before: counts UHC=$NUHC0 injected=$NINJECTED0 Bridge=$NBR0 Claim=$NCL0  hash_chain=${HCV0:0:16}…"
+say "before: digests uhc=${UHC0:0:12} inj=${INJ0:0:12} bridge=${BR0:0:12} claim=${CL0:0:12}"
+[[ "$NUHC0" -ge 1 && "$NINJECTED0" -ge 1 ]] || fail "fixture too thin (UHC=$NUHC0 inj=$NINJECTED0) — run traffic first"
```
mandrigin: **Blocking test-claim gap:** this full-store gate still accepts a fixture with zero BridgeEvents and zero ClaimEvents; only GER/UHC is required non-zero. The later Bridge/Claim equality assertions then pass vacuously, so the advertised full event-fidelity recovery can be green without testing either path. Require `NBR0 >= 1` and `NCL0 >= 1` here (alongside GER), and report all four counts in the thin-fixture failure.

scripts/aggkit-preserve-heal.sh
mandrigin: This is improved but still blocking. Poison removal is unchecked under set -uo: rm -f may fail after archive ownership is preserved, and the script never asserts the main/WAL/SHM files are absent before restoring. --archive requests UID/GID preservation, but diff -r verifies only names/content, not ownership, modes, or writeability. The health gate still accepts any fresh non-fatal retry/error logs and never proves the original wedge cleared or an AggKit component progressed; compose start failure also does not stop a potentially partially-started service. Require checked poison deletion, an ownership/mode manifest or writeability proof, and a wedge/component-specific progress check; every post-start failure must stop the service. Keeping open.

src/service_admin.rs
mandrigin: The two reported mutation paths are now safely fixed: requested-faucet by-ID conflicts fail before the bridge preflight, and stale local A versus chain B fails closed. One analogous success-with-missing-state path remains: with no local origin row and the bridge already bound to B, the later OriginBoundToOther arm still returns Ok(B) without persisting B. The registry reconciler treats exactly that missing on-chain-faucet row as anomalous and halts after its grace window. Rebuild B authoritatively or fail with restore/repair instructions instead of returning success. Please also pin these admin branches with targeted tests. Keeping open.

scripts/verify-event-completeness.sh
mandrigin: The new identity check does not yet model production. Runtime BridgeEvent hashes are derived from bare hex(details_commitment) (restore.rs around project_b2agg_note), while this verifier and its fixture derive from NoteId with and without 0x. Those values differ, so a real wrongly-emitted reclaimed log is not forbidden and can still substitute same-block. Obtain each reclaimed note details commitment and derive the exact runtime hash; require complete derivation output. A symmetric deferred substitution also remains: wrong event for deferred D plus missing legitimate L in the same block is reduced to one missing, then D is reclassified deferred and missing becomes zero. Match deferred candidates by identity too and add that fixture. Keeping open.

scripts/e2e-full-db-loss-recovery.sh
mandrigin: Addressed in 82fb11b. The destructive fixture now requires UHC, injected GER, BridgeEvent, and ClaimEvent counts all to be nonzero and reports all four on failure. Resolving.

scripts/__pycache__/lib-verify-completeness.cpython-312.pyc
mandrigin, line 1: Generated Python bytecode must not be committed. Remove this platform/version-specific `.pyc` and add `__pycache__/` / `*.pyc` to `.gitignore` so running the verifier tests cannot dirty future branches.

scripts/test-verify-completeness-substitution.sh
```diff
@@ -0,0 +1,128 @@
+#!/usr/bin/env bash
```
mandrigin, line 1: **Blocking test integration:** this regression is not invoked by `make test-scripts`, which is what the GitHub check runs. CI therefore stays green even if this test fails. Wire it into that target using the release-profile binary produced by `make build` (for example `TOOL_BIN=target/release/bridge-out-tool bash scripts/test-verify-completeness-substitution.sh`) and keep its shell/Python syntax covered.

scripts/aggkit-preserve-heal.sh
mandrigin: Improved substantially in d8a9720/9fe0d77: poison removal, staging/content checks, archive copy, restart detection, and failure shutdown are now present. Two release-safety gaps remain. The health gate still accepts any INFO line as progress, while the caller passes only a generic wedge phrase; 25 seconds without that text does not prove the exact lost transaction advanced and cannot re-wedge. Pass the exact tx/nonce (available to the watchdog) and require durable component progress or an equivalent exact wedge-state check. Also make ownership validation itself fail-closed: include directories plus numeric UID/GID/modes in checked manifests, rather than file-only find output inside process substitution where producer failure can be invisible. Keeping open.

src/service_admin.rs
mandrigin: The persistence/read-back mechanics are fixed, but the new no-local-row OriginBoundToOther branch now silently adopts on-chain faucet B and returns B even though this authenticated call requested faucet F. That conflicts with FaucetRegistryReconciler security policy: an unknown bridge faucet is a possible external admin-key use and only verified --restore is sanctioned to import it. Auto-adopting here launders that tripwire and changes endpoint intent. Fail before mutation with the authoritative B/origin in the error; the operator can explicitly retry for B or run verified restore. Add an endpoint regression for this branch. Keeping open.

scripts/verify-event-completeness.sh
mandrigin: Correction after checking the exact dependency locked by this PR: head 9fe0d775 pins miden-client and miden-client-sqlite-store 0.16.0-alpha.1. Its authoritative store.sql declares input_notes.note_id and details_commitment as BLOB, and serialize_input_note writes both with to_bytes(). My earlier TEXT claim came from a local 0.15 store and was wrong; the BLOB fixture and hex(...) lookup are correct for this PR.

One fail-closed blocker remains in this verifier path: if commitment/hash resolution is unavailable or incomplete, reclaimed rows are added back to EXPECTED, so a wrongly emitted reclaimed BridgeEvent can satisfy that artificial expectation instead of failing the gate. Also take a consistent SQLite snapshot rather than byte-copying only a live main file; if sidecars are present they must be included, and even rollback-journal mode needs snapshot consistency. Make unresolved identity an explicit verifier failure and add that regression. Keeping open.

scripts/__pycache__/lib-verify-completeness.cpython-312.pyc
mandrigin: Addressed in d8a9720: the generated pyc was removed and __pycache__/ plus *.pyc are now ignored. Resolving.

scripts/test-verify-completeness-substitution.sh
mandrigin: Addressed in d8a9720: the substitution regression and shell/Python syntax checks are now wired into make test-scripts with target/release/bridge-out-tool. The GitHub check exercised the target successfully. Resolving.

scripts/aggkit-preserve-heal.sh
mandrigin: The checked directory/file ownership manifests are now correctly addressed. The exact-tx health check is still not a durable success proof and can reject the successful path: it fails on ANY fresh line containing WEDGE_TX, while AggKit success logs include the submitted transaction ID. Conversely, 25 seconds with no mention plus generic INFO can pass before the same transaction retries and re-wedges. Match the tx only when paired with the wedge error, and require a positive exact outcome such as the proxy durably knowing that tx / a success-specific exact-tx transition. Also default WEDGE_TX to the tx discovered by the standalone non-FORCE path. Keeping open.

src/service_admin.rs
mandrigin: Addressed in 7c97a8f: OriginBoundToOther now fails before any mutation, names the authoritative and requested faucets, and requires explicit operator review/verified restore instead of silently adopting the unknown binding. This preserves the registry security tripwire. The branch is direct and the policy helper is pinned; accepting the endpoint-harness gap as non-blocking. Resolving.

scripts/verify-event-completeness.sh
mandrigin: The unresolved-reclaim false green is fixed in 7c97a8f and the missing-identity regression is good. Two release-gate issues remain. First, the pinned 0.16 schema is BLOB, but this push replaced the production-shaped BLOB fixture with TEXT-only coverage based on my corrected 0.15 observation. Keep BLOB as the primary fixture; old TEXT-store compatibility is out of scope and the encoding-agnostic branch is unnecessary under KISS/YAGNI (or at least test both). Second, copying live main/WAL/SHM sequentially is not a point-in-time SQLite snapshot; this store is normally rollback-journal mode, and a write/checkpoint can interleave. Use SQLite backup or pause with guaranteed unpause and copy the main plus any journal/WAL/SHM sidecars. Keeping open.

scripts/aggkit-preserve-heal.sh
mandrigin: Addressed in 812728a: the exact transaction is considered re-wedged only when paired with the wedge error; both chaos and standalone paths supply WEDGE_TX; and success now requires the proxy transactions table to durably admit that exact resent transaction while re-wedge detection remains active. Resolving.

scripts/verify-event-completeness.sh
mandrigin: The production BLOB fixture and point-in-time paused copy are substantively fixed in 812728a. One small operational blocker was introduced by the pause wrapper: cleanup unconditionally runs docker unpause even when this script did not acquire the pause. If the proxy was already intentionally paused, docker pause fails, CPAUSED stays 0, and EXIT still removes someone else’s pause. The normal unpause result is also ignored, so the script may continue into RPC verification with the proxy still paused. Initialize/track pause ownership before installing the trap; cleanup should unpause only when CPAUSED=1, set it to 0 after successful release, and abort if the normal unpause fails. Keeping open.

mandrigin: Accepted as non-blocking for this PR: the verifier now fails closed, the production BLOB path is covered, and the snapshot is frozen during copy. Pause-ownership cleanup can be hardened separately. Resolving.

IvanBelyakoff (approved):

antdanchenko (approved):

PR #103, opened 2026-07-01
PRST-4039: separate L1 and L2 bridge addresses in bridge-autoclaim

iszubok:
## Problem
`bridge-autoclaim` used a single `BRIDGE_ADDRESS` for **both** the L2 `eth_getLogs` discovery filter *and* the L1 `claimAsset`/`isClaimed` target, assuming the canonical CDK shared-address deploy. On the Miden outpost the addresses differ:

| | address |
|---|---|
| L2 bridge (proxy emits BridgeEvent) | `0xc8cbeb…` |
| L1 bridge (claim target) | `0x1348947e…` |

So discovery scanned L2 for the L1 address, found nothing, and never claimed — a silent empty poll loop (only the startup banner, sponsor nonce 0, `isClaimed` false). Proven against the live proxy: `getLogs` for the L1 address → `[]`; for the L2 address → the BridgeEvent.

## Fix
Replace `--bridge-address` / `BRIDGE_ADDRESS` with two explicit flags:
- `--l1-bridge-address` / `L1_BRIDGE_ADDRESS` — `claimAsset` / `isClaimed` on L1
- `--l2-bridge-address` / `L2_BRIDGE_ADDRESS` — `BridgeEvent` `getLogs` filter on L2

`discover()` uses the L2 address; `is_claimed`/`simulate`/`submit` use the L1 address; the startup banner logs both. README updated. On a canonical shared-address deploy, set both flags to the same value.

Builds clean; 18 claimer unit tests pass.

## Deploy note
Needs a new image (0.15.3). The deployment's `BRIDGE_ADDRESS` env must be replaced with `L1_BRIDGE_ADDRESS=0x1348947e…` + `L2_BRIDGE_ADDRESS=0xc8cbeb…` in lockstep (both flags are required). Unblocks PRST-4026.

Closes PRST-4039.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #186, opened 2026-09-07
fix(#185): unresolvable-destination claim reverts with no ClaimEvent (EVM semantics)

### LINKED ISSUE #184: A claim to an unresolvable destination stops AggLayer certificate settlement from that block onward (RD-860 swallow path × aggsender)
Found by the #167 e2e battery (PR #176), bisected on scratch stacks with `scripts/e2e-settlement-bisect.sh` (commit 38173a0 on `refactor/0.16-167-fixed-tip-catchup`). **Not a #167 regression**: the RD-860 path predates it. The battery's fix was ordering-only (the #146 test moved to the end of the suite); the product interaction is open.

## Mechanism

1. A `claimAsset` whose destination cannot be resolved lands on the RD-860 swallow path (`src/service_send_raw_txn.rs` ~L3004): the claim is recorded in `unclaimable_claims` and a **synthetic ClaimEvent is emitted** "so aggkit stops retrying".
2. aggkit's aggsender reads that event as a claim with an unclaim and **cuts every certificate at the block before it, permanently**:
   ```
   WARN flows/flow_base.go:670  found claim with unclaim after later unfinalized claim at block 37, cutting certificate at block 36
   INFO flows/builder_flow_pp.go:103  PPFlow - no bridges or claims found for range: 1 - 36, so no certificate will be built
   ```
   repeated every epoch, indefinitely. The L2 bridge syncer stays healthy (it ingested the later bridge-out, "block 94 processed with 1 events"); aggsender simply refuses to certify at or beyond the poisoned block.
3. Proof of the link: the `unclaimable_claims` row (global_index 0x1, reason `unresolvable_destination`) carries the same eth tx hash as the block-37 ClaimEvent.

## Observed impact

In the e2e suite, the #146 future-nonce test (which uses an unresolvable-destination `claimAsset` as its park/promote vehicle) ran before the L2→L1 tier, and every certificate-dependent tier after it timed out at settlement (900s) — 2 of 3 hardened battery iterations, while the identical bridge-out passed standalone every time.

On a production chain, **one** claim to an unresolvable destination would stop that network's certificate settlement from that block onward until manual intervention.

## Questions to settle

- Is emitting a synthetic ClaimEvent for a swallowed claim the right contract with aggkit at all, or should the swallow path emit nothing (aggkit's retry cost vs. a permanently cut certificate)?
- If an event must be emitted, what shape does aggsender accept without treating it as "claim with unclaim"?
- Operator recovery today: how does a network unwedge (delete the row and the event? re-project?) — needs a runbook entry until fixed.

Repro: `scripts/e2e-settlement-bisect.sh` (arm C: future-nonce + bridge-out on a fresh stack).

https://claude.ai/code/session_01JGoAzmPkjYq85iADpZodwm

### LINKED ISSUE #185: Adopt EVM revert semantics for unreachable-destination claims: don't emit a synthetic ClaimEvent (unifies #103 and #184)
Design decision surfaced by the #167 restore validation (PR #176) and confirmed with live evidence. It unifies two filed symptoms into one root cause.

## The root cause

A `claimAsset` to a destination with no Miden `AccountId` (unresolvable) takes the RD-860 swallow path: it emits a **synthetic ClaimEvent at admission** so aggkit stops retrying, records the claim in `unclaimable_claims` (named the source of truth *because the event misrepresents the flow*), and creates **no Miden note** — there is no account to mint a P2ID to. No funds move on Miden.

That synthetic event is the anomaly, and it breaks things at both ends:

- **#184 (settlement):** with the event present, aggsender reads it as a claim-with-unclaim and cuts every certificate at that block, permanently (`found claim with unclaim … cutting certificate at block N`).
- **#103 (restore):** `--restore` rebuilds history from authoritative Miden note state, so an event with no on-chain source cannot be replayed and is dropped.

## Evidence that the EVM-faithful behaviour is "no event"

Verified on the live restored stack (full report in the #167 e2e results):

- On EVM, a reverted/failed `claimAsset` emits **no log**, and a claim that simply does not exist is ordinary. The restored history — no event, `unclaimable_claims` as the record — is exactly that shape, and it is the **truthful** one: nothing moved on Miden.
- **Absent is strictly safer:** post-restore, aggkit showed 0 `inconsistent` lines, 0 `cutting certificate` lines, certificate range advancing; `logIndex` integrity intact (both dropped events were alone in their block at index 0, no survivor renumbered). A re-driven unresolvable claim is swallowed and re-recorded — a self-healing loop, not a wedge.

| | certification | restore fidelity |
|---|---|---|
| event **present** (today) | wedges (#184) | unreproducible (#103) |
| event **absent** | healthy | faithful by construction |

## The ask

Stop emitting a ClaimEvent for a swallowed claim, so live history matches restored history and both match EVM revert semantics. The one prerequisite, and the reason this is a design decision not a one-line change: RD-860's retry pressure is **real and Miden-specific** (an arbitrary Ethereum address has no `AccountId`), so the emission needs replacing with a **non-asserting** way to tell the consumer the claim is unclaimable — one that suppresses retries **without telling aggkit a claim occurred**. Removing the emission with no replacement re-opens the retry loop.

Closes #103 and #184 once landed (both are this defect seen from one end).

https://claude.ai/code/session_01JGoAzmPkjYq85iADpZodwm

mandrigin:
Adopts EVM revert semantics for a claimAsset whose destination has no Miden AccountId: record the unclaimable entry, accept-and-**revert** (status 0x0, no logs), emit **no** ClaimEvent. Retry suppression moves to `isClaimed(globalIndex)` reading the `unclaimable_claims` record as claimed.

**Why** — the synthetic ClaimEvent was the shared root cause of:
- **#103**: unreproducible on `--restore` (no Miden note to replay from).
- **#184**: read as a claim-with-unclaim, permanently cutting AggLayer certificate settlement at its block.

**Verified safe (source-checked):**
- aggsender builds a certificate's imported exits from synced Claim rows and **never reads isClaimed** (grep of aggkit `aggsender/`,`agglayer/`), so `isClaimed=true` without an event enters no certificate.
- The claim index lives **in the note** (`ProofData.global_index`) and `depositCount` is a bridge-out-only counter, so **no index shifts on recovery**.
- The claim submitter stops retrying: on revert `claimtxman/monitortxs.go` `checkIfClaimed`→`isClaimed`→true→Confirmed; and `GetPendingDepositsToClaim` already excludes any deposit with a `monitored_txs` row.

Live and restored history now agree (neither emits the phantom event). Funds remain on L1 in `unclaimable_claims` pending operator rescue.

**Tests:** no-ClaimEvent + reverted receipt + unclaimable record; `isClaimed` true from an unclaimable record; two swallow-route stub tests updated. 599 lib tests green, clippy clean.

Not yet done: full live e2e on the remote box (next), and describing the new `claim_unclaimable_reverted_total` metric.

Closes #185. Closes #184. Closes #103.

https://claude.ai/code/session_01JGoAzmPkjYq85iADpZodwm

### DIFF
--- a/REPORT.md
+++ b/REPORT.md
@@ -1,0 +1,205 @@
+# PR #186 / issue #185 — live validation report
+
+**Branch** `fix/185-unresolvable-claim-evm-revert`
+**Base** `origin/main` @ `5934507` (branch was 0 behind / 2 ahead at start — **no rebase was needed**)
+**Stack** node `v0.16.0-rc.5` (`461ac961951c`), stock client rc.5, `zkevm-bridge-service:v0.6.4-RC2-pendingbridges`
+**Evidence root** `e2e-results/185-20260907T205206Z/`
+**`gh` is UNAUTHENTICATED on this box** — the PR summary could not be posted; paste this file into #186 by hand.
+
+> Note on the task premise: I could not find #176 anywhere in `origin/main`'s history
+> (`git log --grep=176`, `git log --oneline | grep 176`). The branch is nonetheless on the
+> current `origin/main` tip.
+
+---
+
+## 1. Static checks + baseline round-trip
+
+| check | result |
+|---|---|
+| `cargo test --lib --bins` | **600 passed, 0 failed, 1 ignored** (599 before my fix added one) |
+| `cargo clippy --all-targets -- -D warnings` | **clean** |
+| `SKIP_STATIC=1 ./run-all.sh` | **rc=0** — `logs/03-run-all-rebuilt.log` |
+| baseline L1→L2 | **PASS** — wallet `0 → 1000` units (`+1000`, expected `+1000`) |
+| baseline L2→L1 | **PASS** — certificate settled on L1, `+5000000000000 wei` autoclaimed |
+
+---
+
+## 2. The #185 behaviour, end to end on a live stack
+
+Driven by a new committed harness script, `scripts/e2e-185-unresolvable-claim-revert.sh`
+(added in `7c149bd`). It bridges a real L1 deposit to a freshly generated EVM address whose
+first byte is non-zero — so it can never satisfy the zero-padding fallback and has no store
+mapping — lets the real claimtxman drive `claimAsset`, and asserts on chain state.
+
+Run of record: `logs/04-185-live.log`, evidence in `185-live/`.
+Deposit `deposit_cnt=2`, `global_index=18446744073709551618` (`0x10000000000000002`),
+destination `0xd1336cc3ad5571621acbe631f08557e6422682b3`,
+claim tx `0x99ea4a267efdd39f511711c67c1b5ed2a487914258111acecd749742c076ab22`.
+
+### 2a — revert semantics: **PASS**
+
+```
+PASS: (a) receipt REVERTED: status=0x0 logs=0 block=112
+PASS: (a) NO ClaimEvent for gi=18446744073709551618 — eth_getLogs=0, synthetic_logs=0
+PASS: (a) nonce advanced past the reverted claim (EVM semantics)   [signer 0x6352…, tx nonce 0x1, account nonce 0x2]
+PASS: (a) claim_unclaimable_reverted_total advanced: 0 -> 1
+      (a) unclaimable_claims rows: 0 -> 1
+```
+
+The no-event check is made twice, from both ends a consumer can look: `eth_getLogs` over the
+whole chain filtered on the ClaimEvent topic, matching the globalIndex in the first data word,
+**and** the proxy's own `synthetic_logs` table. Both zero.
+
+### 2b — `isClaimed(globalIndex)` is the retry-suppression signal: **PASS**
+
+```
+PASS: (b) isClaimed(2,0) = TRUE from the unclaimable record, with NO ClaimEvent behind it
+```
+
+`eth_call` to the bridge address, selector `0xcc461632`, returns ABI true.
+
+### 2c — the claim submitter stops re-driving: **FAILED FIRST, then PASS after a product fix**
+
+This is the one that did not hold, and the reason is worth reading in full.
+
+**First run, `logs/02-185-live.log` + `185-live-attempt1-prefix-estimategas/`:**
+
+```
+(c) monitored_txs status=<none> terminal=0; reverted counter 0 -> 2 -> 10 (total attempts: 10)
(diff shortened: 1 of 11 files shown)


mandrigin:
## Live validation on the remote stack — #185 behaviour PASSES end to end

Node `v0.16.0-rc.5`, stock client rc.5, `zkevm-bridge-service:v0.6.4-RC2-pendingbridges`. Full report committed at `REPORT.md` on the branch.

| Check | Result |
|---|---|
| Static (`cargo test`/clippy) + baseline L1↔L2 | PASS |
| **2a** revert semantics | **PASS** — reverted receipt (status 0x0), NO ClaimEvent (`eth_getLogs`=0, `synthetic_logs`=0), nonce advanced, `claim_unclaimable_reverted_total` 0→1 |
| **2b** `isClaimed(gi)` = retry signal | **PASS** — `isClaimed(2,0)=TRUE` from the unclaimable record with no ClaimEvent behind it |
| **2c** submitter stops re-driving | **PASS after a product fix** — first run the submitter re-drove **10×** because only `isClaimed` learned the fact, not `eth_estimateGas`; fixed in `40d6934` (estimateGas now returns `AlreadyClaimed` for an unclaimable gi); re-run: `monitored_txs` terminal after **1** attempt |
| **2d** certificates keep settling (**#184**) | **PASS** — a normal bridge-out after the unresolvable claim settled at a strictly greater height; **zero** `cutting certificate` lines |
| **2e** restore fidelity (**#103**) | **PASS** — full-DB-loss restore byte-identical; 9 logs → 9, zero exempt, zero unexplained. Before #185 this same run was 10→9 with one EXEMPT drop |

**One real product gap found and fixed** by the live test: `eth_estimateGas(claimAsset)` didn't read the `unclaimable_claims` record, so `checkIfClaimed` was never reached and the submitter re-drove. `40d6934` closes it — both `eth_call isClaimed` and `eth_estimateGas` now read the same durable record. Two harness fixes as well (root-owned isolated store; the completeness verifier still demanded the phantom event).

**By-design floor (not a bug):** the *first* claim attempt still reverts once and consumes one nonce — its `eth_estimateGas` runs before the `unclaimable_claims` row exists. That is exactly EVM behaviour (the first attempt is the one that reverts).

**Still to finish:** the full growing-chain regression battery (proving #185 broke nothing across every scenario) — the session was OOM-killed mid-run; re-running now.

IvanBelyakoff (approved):

Issue #193, opened 2026-09-09
Runbook/DR: recover a chain wedged on an orphaned faucet generation (fail-closed halt is correct; root cause #196)


mandrigin:
Found by the #186 growing-chain battery (gate ×3). **Not a #185/#186 regression** — it's a B2AGG/faucet recovery gap in the #167 family.

## Symptom
After g1's `full-db-loss-recovery-postchaos` drill, the synthetic tip froze permanently at block 4577 and every subsequent target (g2 6/41, g3) failed the liveness check ("eth_blockNumber did not advance"). Restarts do not clear it.

## Root cause
The projector halted fail-closed:

```
projector halted (fail-closed): note 0x8c15b37d…43654ddf1 (LET index 38) is
reserved but its BridgeEvent was never emitted (unrecoverable metadata /
quarantined leaf). Sealing past it would leave a getLogs gap that halts aggkit
bridgesync.
```

The underlying reason (src/projection.rs:485):

```
restore: B2AGG unknown faucet: unknown faucet ID 0xfe545e3a625e30111a06f1da86aa2f:
not found in faucet registry. Register the faucet via admin_registerFaucet or
bridge a claim first.
Error: restore: bridge-out 0x5ebae939…30a8c7bf at block 4604 is minted by faucet
0xfe545e3a625e30111a06f1da86aa2f, which has not [been rebuilt]
```

Faucet `0xfe545e3a…` was **live on-chain** (it minted bridge-out `0x5ebae939` at block 4604) but is **absent from `faucet_registry` (count 0)** after the DB wipe, and `faucet_bootstrap::rebuild_missing_faucet_identities` did **not** rebuild it during `--restore` (it rebuilt other faucets — src/faucet_bootstrap.rs:105 INFO lines — but missed this one). The B2AGG leaf minted by it is therefore unresolvable → quarantined → the projector fail-closes and the whole synthetic chain stops.

## Why it's a real bug, not a chaos artifact
`scripts/chaos-garbo.sh:20-24` documents that the unknown-faucet class is **intentionally not automated and cannot be produced with the shipped bridge-out-tool** — so this faucet is a genuine mid-run faucet whose recovery failed, not an injected adversarial input.

## Impact
[...]

PR #194, opened 2026-09-09
test(harness): bullet-proof e2e/N=30 — classify HEALTHY/INFRA/WEDGED + nuclear reset

mandrigin:
Refs #193. Based on `main`.

A projector fail-closed halt froze the synthetic tip, but the battery only recognized nonce gaps. Every later target waited out its timeout, producing hours of failures without identifying the shared cause. Unhealthy or partially reset containers caused similar misleading runs.

This change adds a shared stack-health classifier for the battery and bridge-loadtest harness. It distinguishes healthy, infrastructure-failed and projector-wedged states; checks base services, proxy RPC/health and an existing L2B overlay; and permits the normal pre-bring-up state when no proxy exists. The battery records the projector halt reason and evidence once, then marks later targets `SKIP-wedged`. A teardown helper supports a fresh run after a broken fixture.

The RD-913 duplicate-BURN check now examines returned rows directly and suppresses psql command tags. Counting output lines misread empty results, and the intermediate correction still interpreted `INSERT 0 0` as a returned row. `psql -q` preserves actual `RETURNING` rows while suppressing that tag. Five mocked classifier cases cover fresh/base/L2B states and infrastructure failures.

This PR detects the #193 halt; the duplicate-faucet prevention is in #198.

The readiness test also handles repair that finishes before HTTP starts. With Web3Signer, custody verification waits behind the initial client sync; the captured claim was repaired at 10:07:02.637, before HTTP bound at 10:07:02.747. Requiring a visible 503 falsely failed that safe startup. The test now requires either an observed 503 with backlog or timestamped evidence that the exact claim was seeded and repaired before serving. It independently probes the live endpoint with a controlled durable-backlog entry (503, then 200 after cleanup), retains byte-for-byte calldata and consumer settlement assertions, normalizes failed curl responses, and parses ANSI tracing fields correctly. Six verdict regression cases cover missing evidence, wrong claims, and repair after binding.

Stack: `main` (`b01ff6d`, v0.16.3) → #194 (`3712a69`) → #197 (`e19bced`) → #198 (`4ca55ba`). Rebased on main on 2026-09-15.

## Validation

Combined-stack SSH results:

| Check | Status |
|---|---|
| Unit/binary tests | 652 passed, 1 ignored |
| Full PostgreSQL module, real DB and all migrations | 52 passed, 2 ignored |
| Script checks and stack-health cases | Passed again on `4ca55ba` in 37s, including six readiness verdict regressions; all 5 classifier cases passed; live health and injected halt-log classification also verified |
| RD-913 restart and duplicate BURN regression | Passed on `8886471` in 182s; real BURN baseline, duplicate assertion executed |
| Full E2E with Web3Signer, remote prover and L2B | Passed in 6,326 seconds (1h 45m 26s) |
| Dedicated fresh-fixture deposit and withdrawal | Passed on `8886471` |
| Full database loss, live baseline | Passed: 8 events retained in served order; all fields except 7 transaction hashes unchanged; no poison; fresh GER and real claim settled |
| Dedicated faucet-identity restore | Passed on `8886471` in 342s |
| Recovery readiness ×3 | Passed on `4ca55ba` in 725s: exact original calldata, live backlog 503→200, consumer re-fetch/resync, three newer certificates confirmed on L1 |
| N=30 | Passed on `4ca55ba` in 2,671s: 11 deposits + 19 withdrawals, 30/30 claimed, zero failures |
| Exact-block completeness | Passed on `4ca55ba` in 23s; N=30 also passed its own exact-block check (22 B2AGG, 22 CLAIM, 33 GER; zero late/missing/extra) |
| Chaos and post-chaos database loss | Chaos cancelled at user request on Sep 15, 12:09 UTC to prioritize urgent #215. No final chaos verdict; post-chaos DB loss not run. Earlier N=30, completeness, full E2E and DB-loss passes remain valid. |
| GitHub CI | All required checks passed on the current restacked heads; optional workflow E2E skipped |

Unit/PostgreSQL/script tests and the completed full E2E ran on `998cd29`. The only subsequent changes are harness corrections in #194: RD-913 psql output handling at `8886471`, then readiness evidence and its regression cases at `4ca55ba`. Production source, Cargo.toml and Cargo.lock are identical; the SSH runner checks this explicitly. Dedicated RD-913, full database loss and faucet identity restore passed on `8886471`. The resumed runner passed scripts, readiness ×3, N=30 and completeness on `4ca55ba`. Its later chaos run was stopped early at user request; it is not counted as a pass. The first readiness attempt is retained as FAIL, with its cause explained in #194; it is not counted as a pass.

The full suite passed concurrent first claims, dynamic ERC-20, faucet-identity restore, foreign-claim isolation, cursor/GER restart durability, private-note restore, L2B transfers and same-address isolation, all three native round trips, public registration concurrency/durability, multi-network metadata recovery after full DB drop, manual-user/sponsor claim races, crash recovery with fresh transfers in both directions, and future-nonce mempool behavior. The crash test reached exactly-once completion and drained pending work; its dedicated orphan-recovery counter did not advance, so the run may have recovered through normal writer retries.

The full-loss RPC snapshots were independently compared as served: strictly increasing, unique `(block, log_index)` pairs; identical first seven fields; zero missing or extra events; exactly seven transaction-hash changes. The recovery procedure resets the surviving consumer nonce state; a fresh injected GER and a settled, note-backed claim provide the final liveness proof. The optional manual-claim allowlist leg was not enabled in `all`.

Logs, metrics and snapshots: `/home/mandrigin/gate-logs/pr-198-stacked-998cd29/` (original full E2E and recovery evidence, including the readiness failure) and `/home/mandrigin/gate-logs/pr-198-stacked-4ca55ba/` (resumed gates) on SSH. Debug writer diagnostics are enabled and continuously captured across container restarts. N=30 and chaos require exact-block completeness (`ALLOW_LATE=0`). #198 remains draft. The interrupted chaos run included supplemental faults during bulk submissions and watchdog recovery attempts; it cannot establish unassisted recovery. Cancellation evidence and a SQL snapshot are preserved in the resumed log directory.


mandrigin:
## Broadened from the projector-only detector into a bullet-proof stack classifier (aa8ec21)

Two real incidents this session, both of which the matrix blamed on innocent targets:
- **INFRA** — a stale/half-reset box left the proxy `unhealthy` and L2B crash-looping; every target failed on `up --wait` → **0/21 matrix**.
- **WEDGED** — the projector fail-closed and the tip froze; 40+ later targets each burned their full timeout.

New `scripts/lib-stack-health.sh`:
- `stack_health` → one read-only verdict + code: `HEALTHY` (0), `INFRA:<reason>` (1 — crash-looping/unhealthy containers, dead RPC), `WEDGED:<reason>` (2 — projector fail-closed / frozen tip). Covers the **web3signer/KMS** overlay when `WITH_WEB3SIGNER`.
- `stack_nuclear_reset` — removes all containers + volumes, because `make e2e-clean-data` does **not** recreate already-running node containers (the exact 0/21 cause).

Wired into:
- **e2e-battery.sh** — the per-target preflight now classifies the whole stack (not just the projector), names the cause once (`stack-infra/stack-wedge-before-<target>`), and skips the rest fast; the one-time genesis wipe does a nuclear reset first.
- **e2e-bridge-loadtest.sh (N=30)** — the settle loop aborts fast with the verdict on a wedge/infra break instead of waiting out the stall window and understating delivery.

This supersedes the projector-only check (same short-circuit, now infra-aware), so it's the single harness-robustness PR.

mandrigin:
Also folded in (4593c41): **fix RD-913's chronic false red.** Its real regression (monitor_burn_serials surviving a proxy restart) passes; the step-5 assertion was the bug — it did `pgquery "INSERT … ON CONFLICT (serial) DO NOTHING RETURNING serial" | wc -l` and required 0, but `pgquery` is `psql -t -A`, whose 0-row result is a trailing empty line, so `wc -l` returned 1 and it cried "ON CONFLICT semantics broken" — while `serial` is the PRIMARY KEY and the DB correctly rejects the duplicate (reproduced live: `INSERT 0 0`). Now it inspects the RETURNING value directly (trimmed). Turns a chronic false-red green; another "harness cries wolf" class removed.

sibuser (approved):

PR #200, opened 2026-09-10
fix: enable TLS for the remote-prover startup probe

### LINKED ISSUE #199: --require-hardening cannot boot with an https:// remote prover URL
## Summary

With `--require-hardening` set and `--miden-prover-url` pointing at an `https://` endpoint, the service fails to start:

```
Error: remote prover unreachable (startup probe)

Caused by:
    0: transport error
    1: Connecting to HTTPS without TLS enabled
```

The prover is reachable; the probe just cannot dial it.

## Cause

The startup probe builds a bare `tonic::transport::Endpoint`:

```rust
let endpoint = Endpoint::from_shared(prover_url.to_string())?
    .timeout(Duration::from_secs(5))
    .connect_timeout(Duration::from_secs(5));
endpoint.connect().await
```

A bare `Endpoint` has no TLS configured, and tonic rejects an `https://` URI outright before any connection is attempted.

The prover client itself connects with native roots (which is why `tonic` is pinned here with `tls-native-roots`), so the probe is stricter than the client it exists to check — it rejects a URL that works fine at run time.

## Why it cannot be configured around

Hardening also *requires* the prover URL to be set, so the three constraints are mutually unsatisfiable for anyone using a TLS-fronted prover:

- hardening requires `--miden-prover-url` — cannot unset it
- the probe cannot dial `https://` — cannot use a TLS URL
- dropping the remote prover means in-process proving, which the docs call out as the OOM cause

This only affects `--require-hardening`, which is why it did not surface earlier: without it the probe is skipped and the prover client connects normally over TLS.

## Suggested fix

Give the probe the same TLS configuration as the client — enable native roots when the scheme is `https`, leaving plaintext URLs unchanged.

PR: #200

## Environment

- v0.16.0
- `--require-hardening` set
- `--miden-prover-url` an `https://` endpoint behind a TLS load balancer

iszubok:
Fixes #199.

The `--require-hardening` startup probe built a bare `tonic::Endpoint`, which has no TLS configured, so tonic refused any `https://` URI with "Connecting to HTTPS without TLS enabled" before a packet was sent. A hardened boot against a TLS-fronted prover therefore failed on a prover that was reachable.

The prover client connects with native roots, so the probe was stricter than the client it exists to check — it rejected configurations that work at run time. And because hardening also requires `--miden-prover-url` to be set, there was no way to configure around it: hardening and a TLS prover were mutually exclusive.

Endpoint construction moves into `prover_probe_endpoint`, which enables native roots when the scheme is `https`. Plaintext URLs are unchanged, so a local/dev prover behaves exactly as before.

## Tests

Two cases in `hardening_tests`, both hermetic — building an endpoint performs no I/O:

- `prover_probe_accepts_https_and_http` — the regression. Fails on `main`.
- `prover_probe_rejects_a_malformed_url` — a bad URL is still a configuration error, so the TLS change does not smuggle one through.

`cargo test --bin miden-agglayer-service hardening_tests` → 25 passed. `cargo fmt --check` clean.

### DIFF
--- a/src/main.rs
+++ b/src/main.rs
@@ -611,6 +611,32 @@ fn parse_bind_addr(s: &str) -> Result<String, String> {
         })
 }
 
+/// Build the endpoint used by the `--require-hardening` remote-prover startup
+/// probe.
+///
+/// TLS has to be configured explicitly for `https://` URLs. A bare `Endpoint`
+/// has no TLS configured, and tonic then refuses the URI outright with
+/// "Connecting to HTTPS without TLS enabled" — so the probe would fail the boot
+/// on a prover that is perfectly reachable. The prover client itself connects
+/// with native roots, so a probe that cannot do the same is stricter than the
+/// client it exists to check, and rejects configurations that work at run time.
+///
+/// Constructing the endpoint performs no I/O; the caller connects.
+fn prover_probe_endpoint(prover_url: &str) -> anyhow::Result<::tonic::transport::Endpoint> {
+    let mut endpoint = ::tonic::transport::Endpoint::from_shared(prover_url.to_string())
+        .context("invalid --miden-prover-url for startup probe")?
+        .timeout(std::time::Duration::from_secs(5))
+        .connect_timeout(std::time::Duration::from_secs(5));
+
+    if endpoint.uri().scheme_str() == Some("https") {
+        endpoint = endpoint
+            .tls_config(::tonic::transport::ClientTlsConfig::new().with_native_roots())
+            .context("failed to enable TLS for the remote prover startup probe")?;
+    }
+
+    Ok(endpoint)
+}
+
 /// Build the JSON-RPC service URL from a validated bind host + port. IPv6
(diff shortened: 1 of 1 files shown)


iszubok:
Retargeted from `main` to `release/0.16`, which I cut from the `v0.16.0` tag (`e7d09ae`) — there wasn't one, and the 0.15 line had `release/0.15`, so this follows that pattern.

`main` is currently identical to `v0.16.0`, so the diff is unchanged either way today. The reason it matters: a 0.16.1 tagged off `main` later would sweep in whatever else lands on `main` first, whereas off `release/0.16` it carries only this.

Needs forward-porting to `main` as well — happy to open that as a separate PR, or it can ride along when `release/0.16` is merged back the way `release/0.15` was.

mandrigin (approved):
Reviewed: enables native-roots TLS for the https prover startup probe (http unchanged), matching the prover client's own connection behavior. Two hermetic tests. Clean, correct fix for #199. All CI green. Approving for the 0.16.1 hotfix (merging to main; release branch is identical).

PR #202, opened 2026-09-10
fix(#201): fee-charging chains — fund only the KMS keys, cascade the rest; fees ON in e2e

### LINKED ISSUE #201: Fee-charging chains: --init fails (empty vaults pay tx fees) — print KMS addresses + amounts, fund KMS keys & cascade, enable fees in e2e
## Problem

On any Miden chain with a non-zero base fee, `--init` cannot deploy the proxy's accounts:

```
INFO  deploying ger_manager account 0x…
Error: transaction execution failed
Caused by: failed to execute transaction kernel program:
  assertion failed: failed to remove the fungible asset from the vault since
  the amount of the asset in the vault is less than the amount to remove
```

This blocks deployment to **every real chain** (testnet/mainnet). It is invisible in our e2e because the dev genesis zeroes the fee.

**Real testnet parameters** (confirmed at genesis block 0 and current tip block 40780):

| | |
|---|---|
| Base fee | **7** |
| Native / fee asset id | `0x18101fa522c174b165efd4f70a0385` |

## Root cause

Protocol 0.16 charges `verification_base_fee` on **every transaction, paid by the executing account from its own vault** in the chain's native fee asset. Miden has no gas-paying EOA: the account itself pays; its signing key only *authorizes*.

- `deploy_account` (`src/init.rs:115`) deploys via `TransactionRequestBuilder::new().build()` — an **empty dummy txn with no input note**, so there is nothing to draw the fee from; the vault must already hold the fee asset → the assertion.
- Our e2e genesis hides this: `fixtures/genesis/02-with-account-files.toml` sets `verification_base_fee = 0`.
- The proxy's `zero_fee_policy_manager` is unrelated — it zeroes what an account charges *others* for consuming its notes, not the fee an account pays to execute its own txns.

## Which accounts pay (key custody ≠ fee vault)

Only **2 accounts hold KMS keys** (`service`, `ger-manager`). But **≥4 accounts need funded vaults**, because the fee is charged to whichever account *executes*, from *that* account's vault — even keyless ones:

| Account | In KMS | Executes / pays | Where |
|---|---|---|---|
| `ger_manager` | ✅ | deploy + **every GER injection** | `init.rs:128`, `ger.rs:193` |
| `service` | ✅ | deploys on first use, **every CLAIM**, faucet mints | `claim.rs:933`, `faucet_ops.rs:301` |
| `bridge` | ❌ keyless (auth'd by service/ger_manager) | network txns hit **its** vault | deploys on first use |
| each faucet | ❌ keyless (bridge-owned) | deploy + mint from **its** vault, one per token | `faucet_ops.rs:119` |

Fees are per-transaction and continuous, so this is ongoing balances, not a one-shot.

## Required fix (urgent)

1. **Enable fees in e2e** — set `verification_base_fee > 0` in the e2e genesis (e.g. `7`, matching testnet) so CI reproduces and gates this. Must land together with the funding fix or the battery goes red.
2. **Know what to fund** — a CLI utility / `--init` flag that prints, *before any deploy*, the **public addresses of the KMS-keyed accounts** (`service`, `ger_manager`) and the **recommended fee-asset amounts** (derived from the chain's live `verification_base_fee` and fee asset id read from the block-header `fee_parameters`, × estimated txn cost × a safety buffer). Addresses are deterministic from the KMS keys.
3. **Fund only the KMS keys; cascade the rest** — the operator funds just `service` + `ger_manager`. Then the proxy has those funded accounts **bootstrap the keyless accounts**: send the bridge and each faucet a fee-asset P2ID note, which they **consume as their deploy transaction** (the consumed note populates the vault *before* the fee is deducted, so the deploy self-funds). Replaces the empty dummy-txn deploy.
4. **Ongoing top-up awareness** — surface per-account fee balances (metrics/log) so operators see vaults draining; `ger_manager` in particular pays on every L1 GER update.

## Design notes

- Keep custody unchanged: still exactly 2 KMS keys. Cascade funding uses the existing RBAC (service administers the bridge; faucets are bridge-owned), so no new keys.
- The fee asset id must always be read from the chain (`fee_faucet_id_from_chain`), never hardcoded — a wrong id fails every fee check.
- A zero-fee chain must keep working unchanged (fee = 0 ⇒ no funding needed).

mandrigin:
Fixes #201.

## Problem
Protocol 0.16 charges `verification_base_fee × (ilog2(cycles)+1)` on **every** transaction, paid by the **executing account from its own vault** in the native fee asset (the live testnet runs base fee 7). A key only authorizes; the account pays. So `--init` on a real chain failed at the first deploy ("amount in the vault is less than the amount to remove"), and nothing told the operator what to fund. Our dev genesis hid it with base fee 0.

## Design — fund only the two KMS accounts; the proxy cascades the rest
1. **`--print-funding`** builds and *persists* `service`/`ger_manager` (random-seeded ids exist only once persisted), reads the chain's fee parameters, writes **`funding.toml`** (addresses, fee asset, `max_fee_per_txn`, recommended `fund_service` / `fund_ger_manager`) and exits without deploying. Re-running re-prints, never rebuilds.
2. The operator sends **two P2IDs** of the fee asset (treasury / faucet operator). `--init` resumes the recorded accounts, waits for the notes (`--funding-wait-secs`), and **deploys each account by consuming its note** — the vault fills before the fee is charged, so the deploy pays for itself. Already-deployed accounts are skipped on restart.
3. **Keyless network accounts (bridge, every faucet)** are built **P2ID-fundable** the way Miden builds its own post-genesis network accounts (the testnet network-monitor's counter): the upstream component set **plus `BasicWallet`**, with `P2idNote` allowlisted at zero fee (`src/network_accounts.rs`, via the public `NetworkAccount::builder`). `service` cascade-sends each a P2ID; its creation tx consumes it. Ongoing notes need no sponsorship (zero per-note policy: the vault pays); top-ups are more P2IDs. Upstream's builders cannot be created post-genesis on a fee-charging chain at all (no `receive_asset`; every accepted note carries a `NetworkAccountTarget` whose creation FPIs into a not-yet-existing target) — proven live, see the comment trail.
4. Upstream's `is_ger_registered` / `try_faucet_from_account` / `owner_account_id` gate on the canonical **code commitment**, which the extra component changes; replicas keep the storage checks and formulas and drop only that gate. MA#4 treats the P2ID root as canonical.
5. **Recipients pay to consume their mint.** The e2e's isolated wallet is funded with the fee asset from the genesis faucet operator when the chain charges fees (`lib-isolated-wallet.sh`); the runbook says the same for real claimants.
6. **Fee-vault monitor** (`FeeVaultMonitor`, `--fee-vault-poll-secs` 60 / `--fee-vault-warn-txns` 32): exports `bridge_fee_vault_balance{account}` and `bridge_fee_vault_txns_left{account}` for `service`, `ger_manager`, the bridge and every faucet (+ `bridge_fee_max_per_txn`), WARNs below the threshold, ERRORs at 0, and logs every "spent since last check" delta. `bridge-out-tool --vault-balance <id>…` gives the same read-only view from the CLI. Measured at base fee 7: network transactions ~40–50 units, signed client transactions the 210 cap; the bridge ran 27 network txs in ten minutes — alert on `bridge_fee_vault_txns_left < 32`, top up from the metric.
7. **Top-ups apply themselves.** A P2ID to `service` / `ger_manager` is consumed by the proxy on its next monitor tick; one to the bridge or a faucet by the **ntx-builder** — the node refuses user-submitted transactions for a deployed network account, and the ntx-builder only sees notes carrying a `NetworkAccountTarget`, so `bridge-out-tool --fund-fee-asset` attaches it when the target is already on chain (a plain P2ID only works pre-deploy, consumed by the creation transaction). Either way the note pays for its own consume: an empty vault recovers without a restart.
8. **Fee-starved work is held, not failed.** With an empty vault the proxy used to fail each claim cleanly (status-0 receipt), and bridge-service's ClaimTxManager resent a new claim every ~2 s until its **hard-coded ten-entry history** was spent, then marked the deposit permanently FAILED (`claimtxman/monitortxs.go`, `maxHistorySize = 10` — not config): a top-up minutes later recovered nothing. The writer worker now recognises the fee-vault abort, keeps the accepted transaction **pending** (the sponsor just waits on "not mined yet"), parks the job and re-dispatches it every 30 s; the first retry after the top-up lands it. Same path for `ger_manager`'s GER inserts. ERROR log + gauge `agglayer_writer_jobs_parked`.
9. **`e2e-fee-exhaustion` (`service` | `ger_manager` | `bridge`)**, in the battery: each scenario starts one account with a tiny budget (`FEE_TXN_BUDGET_*`), runs deposits until it is dry, asserts the stall is visible (`bridge_fee_vault_txns_left{account}=0`, the monitor's ERROR, the flow stuck), tops the account up the runbook's way, and asserts a new deposit completes **and the deposits stuck while dry land too**. All three pass end to end on the shared box (service, ger_manager at `cef69d4`; bridge at `b08f05a`). The target hands back a default-budget stack so later battery targets never inherit a starved chain.
10. Fee-charging fixes for the test tooling: `fee-funder` stays alive (compose `--wait` treated its clean exit as failure, breaking `make e2e-up`) and re-funds a new manifest; native-faucet / foreign-bridge deployments (permissionless-faucet, claim-provenance, chaos-garbo) pay their fees via the shared deploy path; the e2e's isolated recipient wallet is fee-funded.
11. e2e genesis now runs **base fee 7** (mirrors the testnet); a `fee-funder` sidecar plays the operator with `bridge-out-tool --fund-fee-asset`.

Sizing (measured on the box): a bridge network transaction costs **~112 units** (120 of them burned a 13,440-unit cascade in 2.5 h), faucet mints ~105, signed client transactions the 210 cap. Production cascade default is therefore 256 (53,760 ≈ 480 bridge transactions), `fund_service` 698,880 at base fee 7; the e2e battery — one unattended half-day chain, ~9 new tokens per `test-e2e` — reserves 4096/4096/1024 and 64 cascades, well inside the operator's 1e9.

Safety: a network account runs only allowlisted note scripts and its tx-script allowlist holds only the expiration script, so `BasicWallet`'s send procedures are unreachable; the P2ID script calls `receive_asset` alone; the node store classifies network notes by the `NetworkAccountTarget` attachment only, so stranger P2IDs are invisible to the ntx-builder (no fee-drain).

## Evidence (live on the shared box, web3signer → real AWS-KMS keys)
- **Base fee 7: `e2e-web3signer.sh` 13/13 PASS** — bound service key adopted, no local secret, L1→L2 deposit completes, signer served the signatures, signing fails with the signer down.
- **Full battery at base fee 7, run `final6` (`e608a20`): 22 PASS** — both drills, `test-e2e` 286/0 with every tier (L2↔L2, Miden-origin ×3, permissionless), l1-to-l2, claim-provenance, claim-watcher ×2, b2agg, autoclaim, restore, cantina 6/10/12/13, GER-decomposition, security, fuzz, rd913, rd940, reconciler-cursor. The one non-fee failure, `e2e-reconciler-private-note`, is a pre-existing flake (identical on fee-0 batteries since 2026-09-05) → #205.
- **`e2e-fee-exhaustion`: all three scenarios PASS standalone** with full self-recovery of the stuck deposits (service, ger_manager `cef69d4`; bridge `b08f05a`).
- **Full battery `final7` (`93c8fc5`, everything above): 30 PASS / 2 FAIL of 32, 08:18–15:27Z, no mid-run top-up** — drills, `test-e2e` 283/0, l1-to-l2, claim-provenance, claim-watchers, b2agg, autoclaim, restore, cantina 6/10/12/13, GER-decomposition, security, fuzz, both reconciler tests, **`e2e-fee-exhaustion` (all three accounts recovered in-battery)**, rd940, L2↔L2, Miden-origin ×3, recovery-readiness, fixture bridge-out, DB-loss, loadtest N=30, event-completeness, post-chaos DB-loss. The two FAILs are harness: `rd913` (#206, counts drift across its restart window; 5 of the last 8 batteries incl. pre-fee-7) and `chaos-soak`, whose product verdicts — legit completeness, garbo containment, post-chaos liveness (self-recovered) — all PASS; only its "chaos actually fired" gate failed because the garbo's foreign-claim injector could not create its foreign bridge (it passed the creator wallet of another store) — fixed in `5101416`/`1c81d70`; **standalone chaos soak on a fresh stack at `0af9c22`: all four verdicts PASS** (completeness, garbo containment, chaos fired with 7 faults incl. the foreign injection, post-chaos liveness). The rd913 and private-note harness fixes (`5692089`) were verified standalone too (#205, #206). Live board: https://claude.ai/code/artifact/cf09b77c-3eaf-4882-87ad-3e8e7e2ac303
- A mid-run runbook top-up on the live chain (operator P2ID → `service`, 2,580,480 units) was swept on the next tick, no restart.

## Operator runbook
`docs/operations/provisioning.md` → "Funding the accounts on a fee-charging chain" (manual procedure, `funding.toml` fields, top-ups, recipients).

## Follow-ups
- #205 — `e2e-reconciler-private-note` Phase 0 funding timeout, pre-existing on fee-0 batteries too.
- #206 — `e2e-rd913-restart-burn-collision` counts monitor rows across a restart without waiting for the chain to settle; FAIL in 5 of the last 8 batteries incl. pre-fee-7 runs. Harness fixes for #205/#206 are on this branch (`5692089`), standalone verification in progress.
- #208 — chaos soak: the mixed loadtest's aggkit readiness gate waited for a settled certificate after a mid-run recreate while aggsender correctly had nothing to certify yet (the exits come from the back-ops after the gate); fixed on this branch (`0af9c22`, readiness also accepts an evaluated range) and verified by the green soak above. What recreates aggkit at that point is still open. Not fee-related.
- #207 — LET cardinality gate halted permanently on an on-chain read of 0 leaves (vs 44 local) after a chaos run was killed mid-storm; not fee-related.
- #203 — Cantina #4 `asset_callbacks` alert is a pre-existing false positive (fires on `main` at fee 0 too; never blocks).
- Upstream `miden-agglayer`: a P2ID-fundable option on the bridge/faucet builders (and non-code-gated readers) so the replicas can go away.
- A restarted `--init` currently rolls a fresh bridge id until it deploys — pin bridge/faucet ids in the manifest.

https://claude.ai/code/session_01JGoAzmPkjYq85iADpZodwm

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -444,8 +444,14 @@ e2e-claim-watcher-synthesis: e2e-claim-watcher ## After watcher happy path, simu
 	$(COMPOSE_ENV) ./scripts/e2e-claim-watcher-synthesis.sh
 
 .PHONY: e2e-claim-provenance
+# Each scenario bakes a starved account into its own chain; in the battery every later
+# target shares whatever chain is up, so the target hands back a default-budget stack,
+# pass or fail — otherwise the whole tail of the run inherits a service with no reserve.
 e2e-fee-exhaustion: ## #201: service, ger_manager and the bridge each run out of fee asset mid-flow and must recover when topped up (fee-charging chain; brings its own stacks)
-	for a in service ger_manager bridge; do $(COMPOSE_ENV) ./scripts/e2e-fee-exhaustion.sh $$a || exit 1; done
+	rc=0; for a in service ger_manager bridge; do $(COMPOSE_ENV) ./scripts/e2e-fee-exhaustion.sh $$a || { rc=1; break; }; done; \
+	echo "e2e-fee-exhaustion: handing back a default-budget stack (rc=$$rc)"; \
+	KEEP_CHAIN=0 $(MAKE) e2e-down >/dev/null 2>&1; KEEP_CHAIN=0 $(MAKE) e2e-clean-data >/dev/null 2>&1; $(MAKE) e2e-up >/dev/null 2>&1 || echo "WARN: default-budget bring-up failed"; \
+	exit $$rc
 
 e2e-claim-provenance: ## Deploy a FOREIGN bridge on the same chain, drive a claim through it, assert zero ClaimEvent leakage (stack must be up)
 	$(COMPOSE_ENV) ./scripts/e2e-claim-provenance.sh
--- a/Makefile
+++ b/Makefile
@@ -96,13 +96,12 @@ test-scripts: ## Syntax-check + run the shell guard test harnesses (no docker ne
 	bash scripts/test-quiesce-predicate.sh
 
 .PHONY: test-e2e
-test-e2e: ## Spin up docker stack, run E2E tests, tear down (fully self-contained)
+test-e2e: ## Spin up the stack WITH the L2B overlay, run ALL E2E tiers (incl. L2<->L2 + Miden-origin), tear down (fully self-contained)
 	@echo "╔══════════════════════════════════════════════════════════════╗"
-	@echo "║  Starting E2E stack (Anvil, Miden node, PG, bridge, aggkit) ║"
+	@echo "║  Starting E2E stack + L2B overlay (rollup #2)                ║"
 	@echo "╚══════════════════════════════════════════════════════════════╝"
-	@$(MAKE) --no-print-directory e2e-clean-data
 	@./scripts/ensure-e2e-secrets.sh
-	$(E2E_COMPOSE) up -d --build --wait
+	@$(MAKE) --no-print-directory e2e-l2l2-up
 	@echo ""
 	@echo "Stack is up — running E2E tests..."
 	@echo ""
@@ -112,7 +111,7 @@ test-e2e: ## Spin up docker stack, run E2E tests, tear down (fully self-containe
 			echo "KEEP_CHAIN=1 — leaving the stack UP (the chain must survive this target)"; \
 		else \
 			echo "Tearing down stack..."; \
-			$(E2E_COMPOSE) down -v; \
+			$(L2L2_COMPOSE) down -v; \
 		fi; \
 		exit $$EXIT_CODE
 
--- a/Makefile
+++ b/Makefile
@@ -412,6 +412,10 @@ e2e-l2l2: ## Run the L2<->L2 group (preflight + forward L2B->Miden + back Miden-
 e2e-loadtest-isolated: e2e-up ## Isolated bridge-out reliability loadtest (N=30, 1 ETH + 9 ERC-20) on a fresh stack
 	$(COMPOSE_ENV) env N=$${N:-30} ./scripts/e2e-bridge-loadtest-isolated.sh
 
+.PHONY: e2e-miden-origin-fresh
+e2e-miden-origin-fresh: e2e-l2l2-up ## Fresh stack + L2B overlay, then the three Miden-origin round-trips
+	$(MAKE) --no-print-directory e2e-miden-origin
+
 .PHONY: e2e-miden-origin
 e2e-miden-origin: ## Miden-originated token round-trips (->L2B, ->L1, permissionless ->L2B). L2B overlay must be up (make e2e-l2l2-up).
 	for v in "DEST=l2b" "DEST=l1" "REGISTER_MODE=permissionless DEST=l2b"; do env $$v $(COMPOSE_ENV) ./scripts/e2e-miden-origin.sh || exit 1; done
--- a/Makefile
+++ b/Makefile
@@ -409,6 +409,10 @@ e2e-l2l2-up: e2e-clean-data gen-l2b-configs ## Bring up base stack + L2B overlay
 e2e-l2l2: ## Run the L2<->L2 group (preflight + forward L2B->Miden + back Miden->L2B + evidence). Stack must be up (make e2e-l2l2-up).
 	$(COMPOSE_ENV) ./scripts/e2e-test.sh l2l2
 
+.PHONY: e2e-miden-origin
+e2e-miden-origin: ## Miden-originated token round-trips (->L2B, ->L1, permissionless ->L2B). L2B overlay must be up (make e2e-l2l2-up).
+	for v in "DEST=l2b" "DEST=l1" "REGISTER_MODE=permissionless DEST=l2b"; do env $$v $(COMPOSE_ENV) ./scripts/e2e-miden-origin.sh || exit 1; done
+
 .PHONY: e2e-recovery-readiness
 e2e-recovery-readiness: e2e-l2l2-up ## #148: fresh stack -> land a claim -> run the DESTRUCTIVE recovery-readiness test 3x (retained-PG + reset-Miden-store). DEDICATED gate, NOT in e2e-test.sh all: it drops bridge_db + force-recreates aggkit, which would degrade later suite tiers.
 	# Land a real ClaimEvent (with calldata) for the recovery test to blank + repair.
--- a/Makefile
+++ b/Makefile
@@ -408,6 +408,10 @@ e2e-l2l2-up: e2e-clean-data gen-l2b-configs ## Bring up base stack + L2B overlay
 e2e-l2l2: ## Run the L2<->L2 group (preflight + forward L2B->Miden + back Miden->L2B + evidence). Stack must be up (make e2e-l2l2-up).
 	$(COMPOSE_ENV) ./scripts/e2e-test.sh l2l2
 
+.PHONY: e2e-loadtest-isolated
+e2e-loadtest-isolated: e2e-up ## Isolated bridge-out reliability loadtest (N=30, 1 ETH + 9 ERC-20) on a fresh stack
+	$(COMPOSE_ENV) env N=$${N:-30} ./scripts/e2e-bridge-loadtest-isolated.sh
+
 .PHONY: e2e-miden-origin
 e2e-miden-origin: ## Miden-originated token round-trips (->L2B, ->L1, permissionless ->L2B). L2B overlay must be up (make e2e-l2l2-up).
 	for v in "DEST=l2b" "DEST=l1" "REGISTER_MODE=permissionless DEST=l2b"; do env $$v $(COMPOSE_ENV) ./scripts/e2e-miden-origin.sh || exit 1; done
--- a/Makefile
+++ b/Makefile
@@ -433,6 +433,9 @@ e2e-claim-watcher-synthesis: e2e-claim-watcher ## After watcher happy path, simu
 	$(COMPOSE_ENV) ./scripts/e2e-claim-watcher-synthesis.sh
 
 .PHONY: e2e-claim-provenance
+e2e-fee-exhaustion: ## #201: service, ger_manager and the bridge each run out of fee asset mid-flow and must recover when topped up (fee-charging chain; brings its own stacks)
+	for a in service ger_manager bridge; do $(COMPOSE_ENV) ./scripts/e2e-fee-exhaustion.sh $$a || exit 1; done
+
 e2e-claim-provenance: ## Deploy a FOREIGN bridge on the same chain, drive a claim through it, assert zero ClaimEvent leakage (stack must be up)
 	$(COMPOSE_ENV) ./scripts/e2e-claim-provenance.sh
 
--- a/docker-compose.e2e.yml
+++ b/docker-compose.e2e.yml
@@ -317,7 +317,10 @@ services:
       FEE_TXN_BUDGET_SERVICE: "${FEE_TXN_BUDGET_SERVICE:-4096}"
       FEE_TXN_BUDGET_GER_MANAGER: "${FEE_TXN_BUDGET_GER_MANAGER:-4096}"
       FEE_TXN_BUDGET_CASCADE: "${FEE_TXN_BUDGET_CASCADE:-1024}"
-      FEE_TXN_BUDGET_CASCADE_TARGETS: "${FEE_TXN_BUDGET_CASCADE_TARGETS:-}"
+      # Every token the battery bridges in costs `service` one cascade; test-e2e
+      # alone creates nine, loadtest and chaos more. 12 reserve cascades ran out
+      # 1.7 h into final6 (service 3.44M → 1.07M), so reserve for the whole run.
+      FEE_TXN_BUDGET_CASCADE_TARGETS: "${FEE_TXN_BUDGET_CASCADE_TARGETS:-64}"
       # Regression requirement: offload ALL proxy tx proving (CLAIM / GER /
       # faucet / bridge-out / init) to an EXTERNAL Miden tx-prover instead of
       # the in-process LocalTransactionProver. Fallback-to-local stays OFF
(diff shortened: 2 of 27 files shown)


mandrigin:
## Live results on the box (web3signer → real AWS-KMS keys, `verification_base_fee = 7`)

**Wallets (KMS-keyed) — works, twice.** `--init` wrote `funding.toml`; the `fee-funder` sent 80,640 / 53,760 units from the genesis faucet operator; `ger_manager` and `service` each **deployed by consuming their note** (fee paid from the note): e.g. `0xcddb7c97…` / `0xca48afbb…` committed. On restart the resume path reused the manifest's accounts and correctly **skipped** re-deploy (`nonce: 1`).

**Network accounts (bridge, faucets) — blocked by the protocol, not by us.** Two attempts, two precise kernel verdicts:

1. P2ID cascade (`caddd5c`): the bridge cannot consume a P2ID —
   `account procedure with procedure root 0x34a56dd1… is not in the account procedure index map` (no `receive_asset`; it is an `AuthNetworkAccount`).
2. Sponsored bootstrap (`d5b85b1`, `[ConstantFeePolicyConfigNote + FeeSponsorshipNote]`): the **funder's** tx dies creating the feature note —
   `before_foreign_load … account 0x2244f1b4… not found at block 21`. Creating any `NetworkAccountTarget` note makes the standards fee manager FPI into the target, and the kernel `validate_active_foreign_account` requires the target's commitment to be **in the account tree**; `ForeignAccount::private` rejects public ids, so local state can't be substituted.

So in 0.16.1: a network account can only be funded via a sponsorship bound to a feature note → that note can only be created once the account is on-chain → deploying needs a funded vault → **post-genesis network accounts cannot be bootstrapped on a fee-charging chain.** (Genesis-created accounts dodge it: genesis writes vaults directly — that's how the node funds its own operator.)

**Ways out**
- Chains we control (e2e, our devnets): a **genesis bundle** — pre-build bridge + faucets (+ optionally service/ger_manager) with the KMS pubkeys as admins and **pre-funded vaults**, and reference them from the genesis TOML (`[[account]] path = …`, `native_faucet = …` so the fee asset id is known up front). `--init` then imports instead of deploying.
- Public fee-charging testnets: needs an upstream primitive (fee-exempt first tx, or `estimate_note_fee` tolerating an absent target, or a creation sponsorship) — to raise with the Miden team with this repro.

Until one of these lands, `make e2e-up` at base fee 7 fails at the bridge bootstrap by design, with exactly this explanation in the log.

mandrigin:
### Follow-up: why the network-account bootstrap is a protocol limit (not client-side), and the official testnet status

**Kernel:** `validate_active_foreign_account` (transaction-core `account.masm`) hard-asserts the foreign account's commitment equals the one in the **account tree** (`ERR_FOREIGN_ACCOUNT_INVALID_COMMITMENT`); there is no new-account branch — the seed-based creation path (`validate_seed`) exists only for the executing account. `ForeignAccount::private` rejects public ids. So a sender can never reference an undeployed network account, and **every** bridge-allowlisted feature note script asserts the `NetworkAccountTarget` attachment (update_ger, remove_ger, deregister, config_agg_bridge, claim, b2agg, plus the standards config notes) — the attachment is what triggers the sender-side `estimate_note_fee` FPI. The only attachment-free note a network account accepts is the `FeeSponsorshipNote`, and its script refuses consumption without its bound feature note. Hence: no feature note ⇒ no sponsorship ⇒ no vault ⇒ no creation tx (`verification_base_fee > 0`). The fix has to be upstream (e.g. a fee-exempt creation transaction, or `estimate_note_fee` treating an empty account-tree leaf as fee 0).

**Official testnet (rpc.testnet.miden.io, node 0.16.0, base fee 7, native faucet `0x18101fa5…` — the same chain as the screenshot):** the status board's 2,948 network transactions are explained by the genesis-born native faucet serving mint requests; Miden's own network-account tutorial (`network_notes_counter_contract.rs`) still pins **miden-client 0.15** (pre-fees) and deploys with a bare first tx, so it is stale on this chain rather than evidence of a working post-genesis bootstrap. Upstream tracks sponsorship *pairing* in the builder (0xMiden/protocol#3710 → PR #3801) but nothing about a new network account's first transaction under fees.

**What works today:** the two KMS-keyed wallets (proven live twice). **What needs a decision:** genesis-declared, pre-funded bridge/faucets for chains we control, and an upstream ask for public fee-charging testnets.

mandrigin:
### Correction: not a protocol limit — an upstream-builder limit, with a verified fix (`ca1daaf`)

The official testnet (node 0.16.0, base fee 7) **does** run a post-genesis network account: the network-monitor's counter (2,989 network transactions on the status board). Its recipe (`bin/network-monitor/src/deploy/counter.rs`) is the missing piece: the network account is built **with a `BasicWallet` and the P2ID script allowlisted at zero fee**, and its creation transaction **consumes a plain P2ID**. A P2ID carries no `NetworkAccountTarget`, so it needs no FPI and can be created for a not-yet-created target — which is exactly why our two wallets always worked and our bridge/faucets (built by the upstream `AggLayerBridge`/`AggLayerFaucet` builders: no `receive_asset`, attachment-only allowlist) never could.

`ca1daaf` replicates both upstream builders component-for-component via the public `NetworkAccount::builder`, with the allowlist extended by `P2idNote` (scheduled at zero fee) and a trailing `BasicWallet` (`src/network_accounts.rs`); the deploy path collapses to the wallet path (`service` cascade-sends the P2ID, the account's first tx consumes it). The sponsorship bootstrap is removed.

Safety, verified in source: a network account runs only allowlisted note scripts and its tx-script allowlist holds only the expiration script, so `BasicWallet`'s send procedures are unreachable; the P2ID script calls `receive_asset` alone. The node store classifies a network note **only by the `NetworkAccountTarget` attachment** (`crates/store/src/db/models/queries/notes.rs`), so a stranger's P2ID is invisible to the ntx-builder — no fee-drain vector, only cosmetic dust.

Live fee-7 KMS run on `ca1daaf` in progress; result to follow. Follow-up worth pushing upstream: a P2ID-fundable option on the `miden-agglayer` builders so the replica can go away.

mandrigin:
### Root cause of "balance stays 0" on a fee-charging chain — controlled diff, identical code (`e5e4c04`), base fee 0 vs 7

| | fee 0 | fee 7 |
|---|---|---|
| Faucet `AssetCallbackFlag` | Disabled | Disabled |
| Cantina #4 `asset_callbacks: expected Enabled, observed Disabled` | fires | fires |
| ntx-builder claim → mint executions | 6 | 6 |
| Mint P2ID reaches the recipient | yes | **yes** (`[consume] notes: 1 committed`) |
| Recipient consume | ok → balance 1000 | **`failed to remove the fungible asset from the vault since the amount of the asset in the vault is less than the amount to remove`** → balance 0 |

**The deposit itself works on this branch** (fee 0: balance 0 → 1000, #147 link checks pass). On a fee-charging chain the **bridged-in recipient's consume of its minted note is a transaction that pays the kernel fee from the recipient's own vault** — a fresh wallet has no fee asset, so the consume aborts. It is the end-user instance of the same bootstrap rule that #201 solved for the proxy's accounts.

Two things this rules out: the branch's account changes (deposit passes at fee 0), and the callback flag (Disabled on both; the `Enabled` experiment `d621683` was wrong and is reverted in `c569550`). The Cantina #4 alert is a **pre-existing false positive** — it fires on `main`'s base-fee-0 batteries too (e.g. `e2e-results/main-20260909T132254Z`) and never blocks; filed separately.

**Fix:** the e2e funds its isolated recipient wallet with the fee asset from the genesis faucet operator when the chain charges fees (the consume already takes all consumable notes in one tx, so the fee P2ID pays for the mint's consume); runbook states that a claimant must hold the native fee asset to consume their mint (standard Miden UX: faucet first). Then the fee-7 KMS e2e is re-run.

iszubok (approved):

